PR #2, opened 2020-03-20
Hundsdorfer problem

elswit:
Time-dependent IVP. with two presets. Good for testing order reduction in RK methods.


Steven-Roberts:
We'll need to do a few optimizations like using `spdiags` and some general code cleanup before merging.

PR #3, opened 2020-06-09
added reaction diffusion splitting for cusp problem

elswit:


src/+otp/+cusp/fdiff.m
```diff
@@ -0,0 +1,15 @@
+function du = fdiff(~, u, epsilon, L)
```
Steven-Roberts, line 1: Let's rename this to `fdiffusion` to be consistent with problem property name

src/+otp/+cusp/jacdiff.m
```diff
@@ -0,0 +1,21 @@
+function j = jacdiff(~, u, epsilon, L)
+
+n = size(u, 1)/3;
+
+y = u(1:n);
+a = u((n + 1):(2*n));
+b = u((2*n + 1):end);
```
Steven-Roberts, line 7: These variables look unused

Steven-Roberts:
Looks ready for a merge now

PR #5, opened 2020-07-02
2 variable  index-1  DAE problem proposed by Ascher et al.

elswit:
This DAE is interesting bc it's the only one in OTP with time-dependent mass matrix.


src/+otp/+ascherlineardae/+presets/Canonical.m
```diff
@@ -0,0 +1,14 @@
+classdef Canonical < otp.ascherlineardae.AscherLinearDAEProblem
+    methods
+        function obj = Canonical
+            tspan = [0.0, 1];
+            
+
```
Steven-Roberts, line 6: Remove extra line of whitespace

src/+otp/+ascherlineardae/+presets/Canonical.m
```diff
@@ -0,0 +1,14 @@
...
+            params.beta = 0.5;
+
+            y0 = [1,0.5]';
```
Steven-Roberts, line 9: Use ";" instead of transposing

src/+otp/+ascherlineardae/AscherLinearDAEProblem.m
```diff
@@ -0,0 +1,36 @@
...
+            validateNewState@otp.Problem(obj, newTimeSpan, newY0, newParameters)
+            
+            otp.utils.StructParser(newParameters) ...
+                .checkField('beta', 'scalar', 'real', 'finite'); 
+        end
+        
+        function sol = internalSolve(obj, varargin)
+            sol = internalSolve@otp.Problem(obj, 'Method', @ode15s,'AbsTol', 1e-10,...
+              'Mass', obj.Rhs.MassMatrix,'MStateDependence','none', ...
+              'InitialSlope', [obj.Parameters.beta-1;0]);
```
Steven-Roberts, line 32: Is it really necessary to have `AbsTol` and `InitialSlope` set?  Also `varargin` is not forwarded as last argument.

Steven-Roberts: We need to add support for `MStateDependence` in `internalSolve` so it doesn't need to be provided here.

src/+otp/+ascherlineardae/AscherLinearDAEProblem.m
```diff
@@ -0,0 +1,36 @@
...
+        end
+    end
+    
+    methods (Access = protected)
+        function onSettingsChanged(obj)
+            beta = obj.Parameters.beta;
+            
+            obj.Rhs = otp.Rhs(@(t, y) otp.ascherlineardae.f(t, y, beta), ...
+                otp.Rhs.FieldNames.Jacobian, @(t, y) otp.ascherlineardae.jac(t, y, beta), ...
+                otp.Rhs.FieldNames.MassMatrix, @(t) otp.ascherlineardae.mass(t,[], beta));
```
Steven-Roberts, line 19: Let's add `MStateDependence` property to RHS

src/+otp/+ascherlineardae/AscherLinearDAEProblem.m
AndreyAPopov: internalSolve needs to support all default Matlab arguments. Anything that odeset() returns basically.

src/+otp/+ascherlineardae/AscherLinearDAEProblem.m
AndreyAPopov: Should it be an Rhs property? Do we support other ones line JPattern?

Steven-Roberts: Yes.  Don't think JPattern is supported yet, but can be added easily

src/+otp/+ascherlineardae/AscherLinearDAEProblem.m
elswit: > Is it really necessary to have `AbsTol` and `InitialSlope` set? Also `varargin` is not forwarded as last argument.

`InitialSlope` was necessary to get ode15s to solve this problem. Without it, it would complain about inconsistent  intial value.  `AbsTol` can be dropped. Will fix passing varargin.

Steven-Roberts:
I think all the issues have been addressed.  I updated `Problem` and `Rhs` to support more `odeset` options and renamed the `MassMatrix` property of `Rhs` to `Mass` for consistency with `odeset`.

AndreyAPopov:
Does this address the InitialSlope issue? I know you did some MassSingular stuff.

Steven-Roberts:
Yeah, `InitialSlope` isn't necessary if `MStateDependence` is set.

AndreyAPopov:
The only problem I see then is operator spacing. In most of the code we have

-1*a + 2*b + 3*c

instead of 

-1 * a + 2 * b + 3 * c

like the code in this branch now has.

should we prefer one over the other?

Steven-Roberts:
I'm not that bothered about the multiplication spacing.  We would have to change many files to make that consistent.

AndreyAPopov:
OK. I tend to agree. I don't think there are any more issues then, correct?

PR #7, opened 2020-07-29
Arenstorf orbit problem

elswit:


src/+otp/+arenstorf/+presets/Canonical.m
```diff
@@ -0,0 +1,14 @@
+classdef Canonical < otp.arenstorf.ArenstorfProblem
+    methods
+        function obj = Canonical
+            params.m1 = 0.012277471;
+            params.m2 = 1-0.012277471;
+
+            y0 = [0.994; 0; 0; -2.00158510637908252240537862224];
+            tspan = [0,20];
```
Steven-Roberts, line 8: Let's use the timespan given in Solving ODEs I

src/+otp/+arenstorf/+presets/Canonical.m
```diff
@@ -0,0 +1,14 @@
+classdef Canonical < otp.arenstorf.ArenstorfProblem
+    methods
+        function obj = Canonical
+            params.m1 = 0.012277471;
+            params.m2 = 1-0.012277471;
```
Steven-Roberts, line 5: Switch to `1 - params.m1`.  `m1` could be an optional argument.

src/+otp/+arenstorf/ArenstorfProblem.m
```diff
@@ -0,0 +1,37 @@
+classdef ArenstorfProblem < otp.Problem
+  % The classic  3-body  problem with parameters from
+  % Reference:  Hairer, Ernst, Syvert Paul Nørsett, and Gerhard Wanner.
+  % Solving Ordinary Differential Equations I: Nonstiff Problems Springer-Verlag, 1987.
+  % CH II, p. 129.
+  methods
+    function obj = ArenstorfProblem(timeSpan, y0, parameters)
+      obj@otp.Problem('Arenstorf Orbit Problem', 4, timeSpan, y0, parameters);
```
Steven-Roberts, line 8: We generally don't include problem in the name

src/+otp/+arenstorf/ArenstorfProblem.m
```diff
@@ -0,0 +1,37 @@
...
+    function validateNewState(obj, newTimeSpan, newY0, newParameters)
+      validateNewState@otp.Problem(obj, newTimeSpan, newY0, newParameters);
+      
+      otp.utils.StructParser(newParameters) ...
+        .checkField('m1', 'scalar', 'real', 'finite', 'positive') ...
+        .checkField('m2', 'scalar', 'real', 'finite', 'positive');
+    end
+    
+    function mov = internalMovie(obj, t, y, varargin)
+      mov = otp.utils.movie.PhasePlaneMovie(obj.Name, @obj.index2label, varargin{:});
```
Steven-Roberts, line 32: `otp.utils.movie.NBodyMovie` might be more appropriate

src/+otp/Problem.m
```diff
@@ -242,8 +242,8 @@ function validateNewState(obj, newTimeSpan, newY0, newParameters)
         end
         
         function mov = internalMovie(obj, t, y, varargin)
-            mov = otp.utils.movie.TrajectoryMovie(obj.Name, @obj.index2label, varargin{:});
-            mov.record(t, y);
+            mov = otp.utils.movie.PhasePlaneMovie(obj.Name, @obj.index2label, varargin{:});
+            mov.record(t, y(1:2,:));
```
Steven-Roberts, line 246: Looks like this code was used for a quick test and got committed accidentally

src/+otp/+arenstorf/jac.m
```diff
@@ -0,0 +1,9 @@
...
+                                                                                                                                                                                                 0,                                                                                                                                                   0,  0, 1;
+ (3*m1*(2*m2 - 2*y(1))*(m2 - y(1)))/(2*((m2 - y(1))^2 + y(2)^2)^(5/2)) - m1/((m2 - y(1))^2 + y(2)^2)^(3/2) - m2/((m1 + y(1))^2 + y(2)^2)^(3/2) + (3*m2*(2*m1 + 2*y(1))*(m1 + y(1)))/(2*((m1 + y(1))^2 + y(2)^2)^(5/2)) + 1,                                                     (3*m2*y(2)*(m1 + y(1)))/((m1 + y(1))^2 + y(2)^2)^(5/2) - (3*m1*y(2)*(m2 - y(1)))/((m2 - y(1))^2 + y(2)^2)^(5/2),  0, 2;
+                                                                                           (3*m2*y(2)*(m1 + y(1)))/((m1 + y(1))^2 + y(2)^2)^(5/2) - (3*m1*y(2)*(2*m2 - 2*y(1)))/(2*((m2 - y(1))^2 + y(2)^2)^(5/2)), (3*m1*y(2)^2)/((m2 - y(1))^2 + y(2)^2)^(5/2) - m1/((m2 - y(1))^2 + y(2)^2)^(3/2) - m2/((m1 + y(1))^2 + y(2)^2)^(3/2) + (3*m2*y(2)^2)/((m1 + y(1))^2 + y(2)^2)^(5/2) + 1, -2, 0];
```
Steven-Roberts, line 7: The spacing here is weird.  Also potential to move repeated subexpressions into auxiliary variables

src/+otp/+arenstorf/f.m
```diff
@@ -0,0 +1,8 @@
+function dy = f(~, y, m1 ,m2)
+
+    D1= ((y(1)+m1)^2+y(2)^2)^(3/2);
+    D2= ((y(1)-m2)^2+y(2)^2)^(3/2);
+    
+    dy = [y(3);y(4);y(1)+2*y(4)-m2/D1*(y(1)+m1)-m1/D2*(y(1)-m2);y(2)-2*y(3)-m2/D1*y(2)-m1/D2*y(2)];
```
Steven-Roberts, line 6: I suspect `[y(3:4); ...` is slightly more efficient than `[y(3);y(4); ...`

src/+otp/Problem.m
elswit: The first two variables of the state are the relevant ones for the movie here. Unlike plot, I don't think Movie functions support indexing( for now). So that why I only pass in the trajectory of y1 and y2.

Steven-Roberts: This is the `Problem` class not `ArenstorfProblem`.  This breaks all the other problem movies and needs to be reverted.

elswit: whoops :)

src/+otp/+arenstorf/+presets/Canonical.m
```diff
@@ -0,0 +1,14 @@
+classdef Canonical < otp.arenstorf.ArenstorfProblem
+    methods
+        function obj = Canonical
+            params.m1 = 0.012277471;
+            params.m2 = 1-0.012277471;
+
+            y0 = [0.994; 0; 0; -2.00158510637908252240537862224];
```
AndreyAPopov, line 7: That is a very oddly specific y0. Matlab tells me that -2.001585106379082e+00 is at 64-bit machine precision equivalent. I am assuming that there is a closed form expression from which you derived this number? Maybe just keep that (even if it has logs and stuff like that).

src/+otp/+arenstorf/ArenstorfProblem.m
```diff
@@ -0,0 +1,37 @@
...
+        otp.Rhs.FieldNames.Jacobian, @(t, y) otp.arenstorf.jac(t, y, m1, m2));
+    end
+    
+    
+    function validateNewState(obj, newTimeSpan, newY0, newParameters)
+      validateNewState@otp.Problem(obj, newTimeSpan, newY0, newParameters);
+      
+      otp.utils.StructParser(newParameters) ...
+        .checkField('m1', 'scalar', 'real', 'finite', 'positive') ...
+        .checkField('m2', 'scalar', 'real', 'finite', 'positive');
```
AndreyAPopov, line 28: It would seem from the definition of m1 and m2, that 0 < m1 < 1, meaning that we can add '@(x) x < 1' as an additional argument to both?

src/+otp/+arenstorf/f.m
```diff
@@ -0,0 +1,8 @@
+function dy = f(~, y, m1 ,m2)
+
+    D1= ((y(1)+m1)^2+y(2)^2)^(3/2);
+    D2= ((y(1)-m2)^2+y(2)^2)^(3/2);
```
AndreyAPopov, line 4: here for readability I would like spaces between the pluses and minuses. in this case it would really help. 

Oh, and a space before the equals sign as well.

src/+otp/Problem.m
AndreyAPopov: Haha.

src/+otp/+arenstorf/+presets/Canonical.m
elswit: It's specific because it's the init condition that makes the solution periodic. I couldn't find the closed-form solution for it. I think we should keep the double prec accuracy.

AndreyAPopov: It is way more than double precision. If this is the value that is specified in the paper/book I agree that we keep it. It also future-proofs us to extended precision or maybe even quad precision with that many digits.

src/+otp/+arenstorf/Arenstorf.m
```diff
@@ -0,0 +1,37 @@
...
+    function obj = Arenstorf(timeSpan, y0, parameters)
+      obj@otp.Problem('Arenstorf Orbit Problem', 4, timeSpan, y0, parameters);
+    end
+  end
+  
+  methods (Access=protected)
+    
+    function onSettingsChanged(obj)
+      m1 = obj.Parameters.m1;
+      m2 = 1- obj.Parameters.m1;
```
Steven-Roberts, line 16: I prefer to keep m2 as a parameter.  Can this be reverted to `m2 = obj.Parameters.m2;`

AndreyAPopov: In the problem context does it make sense for m1 and m2 to not add up to 1?

AndreyAPopov:
I don't see any problems. Good job!

Steven-Roberts:
Looks ready to merge

PR #9, opened 2020-09-05
Kuramoto Sivashinky 1D PDE

AndreyAPopov:
Simple 1D Chaotic test problem that is significantly better understood than Lorenz '96

### DIFF
--- a/src/+otp/+kuramotosivashinsky/+presets/Canonical.m
+++ b/src/+otp/+kuramotosivashinsky/+presets/Canonical.m
@@ -1,0 +1,29 @@
+classdef Canonical < otp.kuramotosivashinsky.KuramotoSivashinskyProblem
+    
+    methods
+        function obj = Canonical(varargin)
+            
+            p = inputParser;
+            addParameter(p, 'Size', 64, @isscalar);
+            addParameter(p, 'L', 25, @isscalar);
+
+            parse(p, varargin{:});
+            
+            s = p.Results;
+            
+            n = s.Size;
+            
+            
+            params.n = n;
+            params.l = s.L;
+            
+            x = linspace(0, 10*pi, n + 1);
+            
+            u0 = 4*cos(x(1:end-1)).';
+
+            tspan = [0, 100];
+            
+            obj = obj@otp.kuramotosivashinsky.KuramotoSivashinskyProblem(tspan, u0, params);
+        end
+    end
+end
--- a/src/+otp/+kuramotosivashinsky/KuramotoSivashinskyProblem.m
+++ b/src/+otp/+kuramotosivashinsky/KuramotoSivashinskyProblem.m
@@ -1,0 +1,32 @@
+classdef KuramotoSivashinskyProblem < otp.Problem
+    
+    methods
+        function obj = KuramotoSivashinskyProblem(timeSpan, y0, parameters)
+            obj@otp.Problem('Kuramoto Sivashinsky Problem', [], timeSpan, y0, parameters);
+        end
+    end
+    
+    methods (Access = protected)
+        function onSettingsChanged(obj)
+            n = obj.Parameters.n;
+            l = obj.Parameters.l;
+            
+            domain = [-l, l];
+            D = otp.utils.pde.D(n, domain, 'C');
+            L = otp.utils.pde.laplacian(n, domain, 1, 'C');
+            
+            obj.Rhs = otp.Rhs(@(t, u) otp.kuramotosivashinsky.f(t, u, D, L), ...
+                otp.Rhs.FieldNames.Jacobian, @(t, u) otp.kuramotosivashinsky.jac(t, u, D, L));
+            
+        end
+        
+        function validateNewState(obj, newTimeSpan, newY0, newParameters)
+            validateNewState@otp.Problem(obj, ...
+                newTimeSpan, newY0, newParameters)
+            
+            otp.utils.StructParser(newParameters) ...
+                .checkField('n', 'scalar', 'integer', 'finite', 'positive') ...
+                .checkField('l', 'scalar', 'integer', 'finite', 'positive');
+        end
+    end
+end
--- a/src/+otp/+kuramotosivashinsky/f.m
+++ b/src/+otp/+kuramotosivashinsky/f.m
@@ -1,0 +1,5 @@
+function du = f(~, u, D, L)
+
+du = - L*(L*u + u) - u.*(D*u);
+
+end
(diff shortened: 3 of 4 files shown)


src/+otp/+kuramotosivashinsky/f.m
```diff
@@ -0,0 +1,5 @@
+function du = f(~, u, D, L)
+
+du = - L*(L*u + u) - u.*(D*u);
```
elswit, line 3: Might add4 spaces here

src/+otp/+kuramotosivashinsky/jac.m
```diff
@@ -0,0 +1,5 @@
+function j = jac(~, u, D, L)
+
+j = -L*L - L - spdiags(u, 0, numel(u), numel(u))*D - spdiags(D*u, 0, numel(u), numel(u));
```
elswit, line 3: and an indent  here

src/+otp/+kuramotosivashinsky/+presets/Canonical.m
```diff
@@ -0,0 +1,29 @@
+classdef Canonical < otp.kuramotosivashinsky.KuramotoSivashinskyProblem
+    
+    methods
+        function obj = Canonical(varargin)
+            
+            p = inputParser;
+            addParameter(p, 'Size', 64, @isscalar);
+            addParameter(p, 'L', 25, @isscalar);
```
Steven-Roberts, line 8: I would remove the validation function and leave it to `validateNewState`

Steven-Roberts: Also, having the size `n` in the parameters is redundant.  There can be inconsistencies with the size of `Y0` and `n`.  I think we should prefer `Problem.NumVars`

src/+otp/+kuramotosivashinsky/jac.m
Steven-Roberts: We should probably precompute L*L to avoid matrix-matrix multiplications every Jac eval

src/+otp/+kuramotosivashinsky/f.m
Steven-Roberts: For the term u u_x, I think it's generally preferable to write as d/dx (u^2) / 2 and discretize as `D u^2 / 2`

src/+otp/+kuramotosivashinsky/javp.m
```diff
@@ -0,0 +1,5 @@
+function jv = javp(~, u, k, k2, k4, v)
```
Steven-Roberts, line 1: The problem parameters are always the last argument.  It should be `(~, u, v, k, k2, k4)`

src/+otp/+kuramotosivashinsky/+presets/Canonical.m
```diff
@@ -0,0 +1,47 @@
...
+            s = p.Results;
+            
+            N = s.Size;
+            L = s.L;
+            
+            params.L = L;
+            
+            h = L/N;
+            
+            x = h*(1:N).';
```
Steven-Roberts, line 35: Why not just use `linspace`

src/+otp/+kuramotosivashinsky/KuramotoSivashinskyProblem.m
```diff
@@ -0,0 +1,57 @@
...
+                @(t, u, v) otp.kuramotosivashinsky.javp(t,u, k, k2, k4, v));
+            
+        end
+        
+        function validateNewState(obj, newTimeSpan, newY0, newParameters)
+            
+            validateNewState@otp.Problem(obj, ...
+                newTimeSpan, newY0, newParameters)
+            
+            assert(mod(numel(newY0), 2) == 0);
```
Steven-Roberts, line 50: We don't use `assert` anywhere else.  Probably should call `error` with a nice error message

src/+otp/+kuramotosivashinsky/KuramotoSivashinskyProblem.m
```diff
@@ -0,0 +1,59 @@
...
+            
+            L = obj.Parameters.L;
+            
+            N = obj.NumVars;
+            
+            div = L/(2*pi);
+            
+            k = (1i*[0:(N/2 - 1), 0, (-N/2 + 1):-1].'/div);
+            k2 = k.^2;
+            k4 = k.^4;
```
Steven-Roberts, line 33: `k2` and `k4` are always used together.  We can consolidate this to `k24 = -k.^2 - k.^4`

src/+otp/+kuramotosivashinsky/jac.m
```diff
@@ -0,0 +1,5 @@
+function j = jac(~, u, k, k2, k4)
+
+j = -diag(k2) - diag(k4) - diag(k)*ifft(fft(diag(ifft(u))).').';
```
Steven-Roberts, line 3: `diag(k)` can be replaced with `.*` to scale rows

src/+otp/+kuramotosivashinsky/KuramotoSivashinskyProblem.m
```diff
@@ -0,0 +1,59 @@
+classdef KuramotoSivashinskyProblem < otp.Problem
+    
+    methods
+        function obj = KuramotoSivashinskyProblem(timeSpan, y0, parameters)
+            obj@otp.Problem('Kuramoto-Sivashinsky', [], timeSpan, y0, parameters);
+        end
+    end
+    
+    methods
+        function soly = solution2real(soly)
```
Steven-Roberts, line 10: To be consistent with Pendulum problem, let's rename to something like `convert2grid`

src/+otp/+kuramotosivashinsky/+presets/Canonical.m
```diff
@@ -0,0 +1,49 @@
...
+            
+            params.L = L;
+            
+            h = L/N;
+            
+            % exclude the left boundary point as it is identical to the
+            % right boundary point
+            x = linspace(h, L, N).';
+            
+            u0 = cos(x/16).*(1+sin(x/16));
```
Steven-Roberts, line 39: Replace hardcoded 16 so i.c. is periodic for different L.  It might be easier to have x in the range 0 to 2 pi.

AndreyAPopov: If I have x in 0 to 2 pi, then I would have to calculate h differently. I will change u0, to be a function of L and use cospi and sinpi. I think that is simpler.

PR #10, opened 2020-09-22
added pendulum problem

rtuggle99:


src/+otp/+pendulum/PendulumProblem.m
```diff
@@ -0,0 +1,50 @@
...
+            obj@otp.Problem('Pendulum', 2, timeSpan, y0, parameters);
+        end
+    end
+    
+    methods (Access = protected)
+        function onSettingsChanged(obj)
+            g = obj.Parameters.g;
+            l = obj.Parameters.l;
+            
+            obj.Rhs = otp.Rhs(@(t, y) otp.pendulum.f(t, y, g, l), ...
```
AndreyAPopov, line 26: I know that this is your first problem, and overwhelming you is not the best thing, but adding derivatives with respect to parameters would be perfect for this simple problem. df/dg and df/dl would be great trivial tests for parameter estimation.

src/+otp/+pendulum/f.m
```diff
@@ -0,0 +1,6 @@
+function dy = f(~, y, g, l)
+dy = [
```
AndreyAPopov, line 2: I'm not sure that I like the spacing here.

AndreyAPopov: Also, support for vectorization is trivial: replace y(1) with y(1, :) and y(2) with y(2, :).

src/+otp/+pendulum/jac.m
```diff
@@ -0,0 +1,10 @@
+function jac = jac(~, y, g, l)
+
+jac = [
+    0, 1;
+    -g/l*cos(y(1)), 0
+    ];
+
+end
+
```
AndreyAPopov, line 9: too many newlines at the end

src/+otp/+pendulum/jac.m
```diff
@@ -0,0 +1,10 @@
+function jac = jac(~, y, g, l)
+
+jac = [
```
AndreyAPopov, line 3: spacing here is weird too, and inconsistent with the spacing in f.

src/+otp/+pendulum/PendulumProblem.m
```diff
@@ -0,0 +1,50 @@
...
+            validateNewState@otp.Problem(obj, newTimeSpan, newY0, newParameters)
+            
+            otp.utils.StructParser(newParameters) ...
+                .checkField('g', 'scalar', 'real', 'finite', 'positive') ...
+                .checkField('l', 'scalar', 'real', 'finite', 'positive');
+        end
+        
+        function label = internalIndex2label(~, index)
+            if index == 1
+                label = 'Theta';
```
AndreyAPopov, line 40: why theta and Omega, do we not want 'Angle' and "Angular Velocity"?

src/+otp/+pendulum/PendulumProblem.m
```diff
@@ -0,0 +1,50 @@
...
+            else
+                label = 'Omega';
+            end
+        end
+        
+        function sol = internalSolve(obj, varargin)
```
AndreyAPopov, line 46: The default is ode15s now, right? do we want to explicitly change it to ode45 for the pendulum? a contrived parameter change to l = 1e-5 might make the problem extremely stiff, right?

src/+otp/+pendulum/PendulumProblem.m
```diff
@@ -0,0 +1,50 @@
...
+    %   This system can be reduced to the following two equations
+    %
+    %   theta' = omega
+    %   omega' = -g/l * sin(theta)
+    %
+    %   where g is the gravitational constant, l is the length of
+    %   the pendulum, and theta is the angle between the pendulum and
+    %   y-axis.
+     
+    methods
```
AndreyAPopov, line 14: Before we merge, we have to have a movie for the pendulum. This is one of the more perfect problems for this.

src/+otp/+pendulum/PendulumProblem.m
Steven-Roberts: Might as well add Jacobian-vector products too.  Maybe even Hessian too

src/+otp/+pendulum/PendulumProblem.m
```diff
@@ -0,0 +1,50 @@
...
+            else
+                label = 'Omega';
+            end
+        end
+        
+        function sol = internalSolve(obj, varargin)
+            sol = internalSolve@otp.Problem(obj, 'Method', @ode45, varargin{:});
+        end
+    end
+end
```
Steven-Roberts, line 50: Add newline at end of file

Steven-Roberts:
Let's add a preset where the pendulum barely completes full rotations

src/+otp/+pendulum/PendulumProblem.m
AndreyAPopov: and adjoint information (jac adj vp, hes asj vp) as well. For a problem as simple as this, we have to.

AndreyAPopov:
If we are brainstorming preset ideas, would a preset where the gravitational constant is not a constant and is time dependent (simulating the pendulum anchor point being jerked around) be useful for any TI-method tests, similar to Prothero-Robinson?

A few more:
1. A trivial stationary preset
2. A preset that is barely moving such that we can compare it to the small angle approximation? (In fact would a small angle Rhs be useful?)

Steven-Roberts:
Honestly a time dependent g doesn't seem that useful.  I think adding a damping term could be nice.

The other two suggestions sound good

AndreyAPopov:
I like the damping term idea. Would a generic forcing function that depends on theta and theta prime be useful, or would we be in danger of approaching oscillators at that point?

AndreyAPopov:
I just thought of another important function that needs to exist: converting to and from cartesian coordinates. If you assume that the pivot is located at (0,0), the angle theta coupled with the length of the pendulum gives you a set of x and y coordinates. The angular velocity additionally gets you dx/dt and dy/dt.

A set of (x,y) coordinates, and (dx/dt,dy/dt) also gives you an over-determined system for the angle and angular velocity.

It would be beneficial to be able to convert between the two representations, especially for testing data assimilation systems.

AndreyAPopov:
I made the mass function vectorized. There is still A LOT to do with the code, but I like that we have n-pendulum this nice. Have we confirmed that it is different from the "Rope" problem form the Bible part 1?

Steven-Roberts:
I think it's possible to cast ROPE into this framework but we would need to support external forcing.

AndreyAPopov:
Rope had an external force only for part of the time frame, essentially to setup the real initial condition. I believe we can just setup a similar initial condition and say that that is good enough.

Steven-Roberts:
The bottom half is dense.  I think the most efficient way to work with the diagonal is `data1(1:(n+1):end)`.  I'll commit an optimized version tonight.

Steven-Roberts:
Some testing shows the `diag(diag(...))` is actually pretty efficient.  More efficient than `data1(1:(n+1):end)`.

AndreyAPopov:
I would assume that for large n it would be less efficient. If it is faster for a reasonable number (128) then I vote keep it. Maybe test for 1024 and if the current version is significantly slower, choose an 'n' in onSettingsChanged, and use the diagonal implementation? If it is not significantly slower, then it doesn't matter, as the Jacobian is dense anyways.

src/+otp/+pendulum/jac.m
```diff
@@ -1,16 +1,12 @@
...
-dada1 = dada1 - diag(sum(dada1,2));
-dada = dada1 + diag(cumulativeMasses * -g .* lengths .* cos(angles));
-jac = [
-    zeros(n), eye(n);
-    dada, dadv
-    ];
+dAccelDVel = -2 * scaledMasses .* velocities .* sin(angles - angles.');
+dAccelDAng = scaledMasses .* (velocities.^2) .* cos(angles - angles.');
+dAccelDAng = dAccelDAng - diag(sum(dAccelDAng, 2) + cumulativeMasses * g .* lengths .* cos(angles));
+jac = [zeros(n), eye(n); dAccelDAng, dAccelDVel];
```
AndreyAPopov, line 9: I think this version resolves all outstanding issues.

AndreyAPopov:
The issue with the jacobian now is creating jacobian and adjoint vector products. For a dense jacobian that is very important.

Steven-Roberts:
Here are the remaining things before we can merge.

- [ ] Finish implementing problem functions like `validateNewState` and `internalIndex2Label`
- [ ] Add function to convert from polar to Cartesian coords
- [ ] JVP

Extras
- [ ] JAVP
- [ ] Form without the mass matrix using the trick from ROPE
- [ ] External forcing

You can edit if there's more

src/+otp/+pendulum/PendulumProblem.m
mod0: here's a tip for you fellows - checkout pre-commit(https://pre-commit.com/) and you can setup hooks that will format your code based on a standard scheme. For python we use something called `black`. There are other formatters for Matlab which will auto-format every code before checkin. Of course you can get fancy and install other pre-commit hooks: linters, test runners etc. Once you set them up, commits will only succeed once code is reformatted, linter does not complain and your test cases pass.

Steven-Roberts: That looks cool. I wasn't able to find a Matlab formatter though. If we find one I'm all for trying it

mod0: Maybe this one: https://github.com/davidvarga/MBeautifier

mod0: You may have to write a bash script to invoke it on each `.m` file in the folder

src/+otp/+pendulum/PendulumProblem.m
```diff
@@ -36,16 +36,18 @@ function onSettingsChanged(obj)
         function validateNewState(obj, newTimeSpan, newY0, newParameters)
             validateNewState@otp.Problem(obj, newTimeSpan, newY0, newParameters)
             
-            %            otp.utils.StructParser(newParameters) ...
-            %                 .checkField('g', 'scalar', 'real', 'finite', 'positive') ...
-            %                 .checkField('lengths', 'vector', 'real', 'finite', 'positive');
+            otp.utils.StructParser(newParameters) ...
+                .checkField('g', 'scalar', 'real', 'finite', 'positive') ...
+                .checkField('lengths', 'vector', 'real', 'finite', 'positive') ...
+                .checkField('masses',  'vector', 'real', 'finite', 'positive');
```
AndreyAPopov, line 42: would vector still work for a scalar?

src/+otp/+pendulum/PendulumProblem.m
```diff
@@ -36,16 +36,18 @@ function onSettingsChanged(obj)
...
-            if index == 1
-                label = 'Theta';
-            else
-                label = 'Omega';
+        function label = internalIndex2label(obj, index)
+            numPendulums = obj.NumVars/2;
+            if index <= numPendulums
+                label = sprintf('\\theta_{%d}', numPendulums);
+            else 
+                label = sprintf('\\omega_{%d}', index - numPendulums);
```
AndreyAPopov, line 50: I like these names, but would 'Angle' and 'Angular velocity' be more descriptive? I'm fine with keeping this.

src/+otp/+pendulum/PendulumProblem.m
Steven-Roberts: Yes.  It uses `isvector` which is true even for scalars.

src/+otp/+pendulum/PendulumProblem.m
Steven-Roberts: The latex displays nicely on plots.  `label2index` is not really used outside of plotting.

Steven-Roberts:
The mass matrix form requires solving a dense matrix, so the non-mass-matrix form is faster.  As far as I can tell, backslash is the fastest way to solve tridiaginal systems in Matlab.

src/+otp/+pendulum/PendulumProblem.m
```diff
@@ -0,0 +1,79 @@
...
+        RhsMass
+    end
+    
+    methods
+        function obj = PendulumProblem(timeSpan, y0, parameters)
+            % Constructs a problem
+            obj@otp.Problem('Pendulum', [], timeSpan, y0, parameters);
+        end
+        
+        function [x, y] = convert2Cartesian(obj, y, includeOrigin)
```
AndreyAPopov, line 12: includeOrigin need a sane default

src/+otp/+pendulum/PendulumProblem.m
```diff
@@ -0,0 +1,79 @@
...
+            
+            numBobs = min(numel(lengths), numel(masses));
+            lengths = lengths(1:numBobs);
+            masses = masses(1:numBobs);
+            cumulativeMasses = cumsum(masses, 'reverse');
+            
+            offDiagScaling = -1 ./ (lengths(1:end-1) .* lengths(2:end) .* masses(1:end-1));
+            cDiag = [1 / (lengths(1)^2 * masses(1)); (masses(1:end-1) + masses(2:end)) ./ (lengths(2:end).^2 .* masses(1:end-1) .* masses(2:end))];
+            
+            obj.Rhs = otp.Rhs(@(t, y) otp.pendulum.f(t, y, lengths, cumulativeMasses, g, offDiagScaling, cDiag));
```
AndreyAPopov, line 40: JVP? If it's too difficult, then meh.

src/+otp/+pendulum/PendulumProblem.m
AndreyAPopov: Also a function to get the angles from cartesian would be nice

src/+otp/+pendulum/+presets/DoublePendulum.m
```diff
@@ -0,0 +1,14 @@
...
+            params.lengths = [1,1];
+            params.masses = [1,1];
+            
+            y0 = [pi/3; 0; 0; 0];
+            tspan = [0 10];
+            
+            obj = obj@otp.pendulum.PendulumProblem(tspan, y0, params);
+        end
+    end
+end
```
AndreyAPopov, line 14: Can we make a simple n pendulum preset?

src/+otp/+pendulum/PendulumProblem.m
Steven-Roberts: includeOrigin already has a default value of false.  A Cartesian to polar would probably have to be a static function since (x,y) points may not be consistent with Parameters.lengths.  For consistency and simplicity, I don't think it's worth adding.

src/+otp/+pendulum/PendulumProblem.m
Steven-Roberts: Jacobian and JVP is complicated.  We can add it later if needed.

src/+otp/+pendulum/+presets/DoublePendulum.m
Steven-Roberts: Yes, will adjust the Canonical preset

AndreyAPopov:
I think you can merge

PR #11, opened 2020-12-16
Lorenz 96 Quadratic ROM

AndreyAPopov:
This is a quadratic reduced order model for the Lorenz '96 equations for use in an upcoming paper.


src/+otp/+lorenz96/f.m
```diff
@@ -1,5 +1,5 @@
 function dy = f(t, y, f)
 
-dy = -y([end, 1:(end - 1)]).*(y([end - 1, end, 1:(end - 2)]) - y([2:end, 1])) - y(1:end) + f(t);
+dy = -y([end, 1:(end - 1)], :).*(y([end - 1, end, 1:(end - 2)], :) - y([2:end, 1], :)) - y + f(t);
```
Steven-Roberts, line 3: Now that it's vectorized, we can add the `Vectorized` field to the Rhs

src/+otp/+quadratic/+presets/L96PODGROM.m
```diff
@@ -0,0 +1,35 @@
+classdef L96PODGROM < otp.quadratic.QuadraticProblem
+    
+    properties
+        Projection
+        Interpolation
+    end
+
+    methods
+        function obj = L96PODGROM(r)
```
Steven-Roberts, line 9: Maybe rename to Lorenz96PodRom to be consistent.  For example, we use Rhs instead of RHS

Steven-Roberts:
I think this can be merged now.

PR #12, opened 2020-12-17
Start reworking the plotting and movies

Steven-Roberts:
- Removed plotState
- Automatically determine if y needs to be transposed for plotting and
movie functions.  If y is square, cols correspond to different times
- Movies support same name value pairs as plot*
- plotPhaseSpace supports multiple lines
- Start optimizing movies by avoiding calls to plot.  Update graphics
objects directly


Steven-Roberts:
Also reverted to MATLAB default colororder instead of rainbow one we enforced before.  Can be easily changed with `'colororder'` name value pair

AndreyAPopov:
we could provide the nice rainbow color order in otp.utils, as the default matlab one has like 7 colors.

Steven-Roberts:
Do you mean as the default or just a separate function like the old `otp.utils.FancyPlot.color`?  You can use `p.plot('colororder', hsv(p.NumVars))` to get rainbow colors.  I'm in favor of FancyPlot not modifying any of the plot defaults unless a name value pair is provided.

AndreyAPopov:
I'm in favor of it not modifying defaults as well, would just be nice to package in an alternative preset with more colors, but I can live without it

Steven-Roberts:
Movies are working correctly now on all problems.  It should be ready to merge once reviewed.

PR #13, opened 2021-04-30
Start the Sanz-Serna Problem

reid-g:
TODO create RHS for the forcing and linear terms


src/+otp/+sanzserna/+presets/Canonical.m
```diff
@@ -0,0 +1,11 @@
...
+    methods
+        function obj = Canonical(m)
+            if nargin < 1
+                m = 32;
+            end
+            x = linspace(1/m + 1, 2, m).';
+            obj = obj@otp.sanzserna.SanzSernaProblem([0, 1], x, struct);
+        end
+    end
+end
```
AndreyAPopov, line 11: Minor thing: new line at the end of file.

src/+otp/+sanzserna/+presets/Canonical.m
```diff
@@ -0,0 +1,11 @@
+classdef Canonical < otp.sanzserna.SanzSernaProblem
+    methods
+        function obj = Canonical(m)
+            if nargin < 1
```
AndreyAPopov, line 4: spacing needs to be consitent. We usually  do a new line around functions.

src/+otp/+sanzserna/+presets/Canonical.m
```diff
@@ -0,0 +1,11 @@
...
+                m = 32;
```
AndreyAPopov, line 5: Most problems use "n" or "N" for the spatial discretization. Why do we use "m" here?

src/+otp/+sanzserna/+presets/Canonical.m
```diff
@@ -0,0 +1,11 @@
...
+            end
+            x = linspace(1/m + 1, 2, m).';
```
AndreyAPopov, line 7: Do we need this to be transposed here?

src/+otp/+sanzserna/SanzSernaProblem.m
```diff
@@ -0,0 +1,27 @@
...
+        function obj = SanzSernaProblem(timeSpan, y0, parameters)
+            obj@otp.Problem('Sanz-Serna', [], timeSpan, y0, parameters);
+        end
+    end
+    
+    methods (Access=protected)
+        
+        function onSettingsChanged(obj)
+            
+            D = spdiags(ones(obj.NumVars, 1) * [obj.NumVars, -obj.NumVars], [-1, 0], obj.NumVars, obj.NumVars);
```
AndreyAPopov, line 12: Here this looks like, but is not (from my understanding) a downwind finite difference, right? The letter "D" implies that it is, however. We might not want to create that impression.

src/+otp/+sanzserna/+presets/Canonical.m
AndreyAPopov: In fact, this should be called "y0". (And it does need to be transposed, yeah.)

AndreyAPopov: Are you sure that you are not mixing this up with the "x" used in the problem? What is defined here should be the initial condition, and be named as such.

src/+otp/+sanzserna/jac.m
```diff
@@ -0,0 +1,4 @@
+function Dd = jac(~, ~, Dd, ~)
```
AndreyAPopov, line 1: Keep the names consistent here

src/+otp/+sanzserna/SanzSernaProblem.m
```diff
@@ -0,0 +1,27 @@
...
+            
+            x = linspace(1/obj.NumVars, 1, obj.NumVars).';
+            
+            obj.Rhs = otp.Rhs(@(t, y) otp.sanzserna.f(t, y, D, x), ...
+                otp.Rhs.FieldNames.Jacobian, @(t, y) otp.sanzserna.jac(t, y, D, x), ...
+                otp.Rhs.FieldNames.JacobianVectorProduct, @(t, y, v) otp.sanzserna.jvp(t, y, v, D, x));
```
AndreyAPopov, line 18: As this is a heavily time-dependent problem, we will want a Jacobian with respect to time as well, ("T" is the literature).  Some implicit methods require it as well from my understanding.

src/+otp/+sanzserna/+presets/Canonical.m
Steven-Roberts: m is what the paper used.  If we rename, something descriptive like `numGridCells` might be best.

src/+otp/+sanzserna/+presets/Canonical.m
Steven-Roberts: Yes let's rename to `y0`.

src/+otp/+sanzserna/SanzSernaProblem.m
Steven-Roberts: This is a first order upwind finite difference matrix.  D for differentiation matrix I guess.

AndreyAPopov: If it is, then should it not be divided by h? Otherwise what is here is wrong

src/+otp/+sanzserna/SanzSernaProblem.m
```diff
@@ -0,0 +1,27 @@
+classdef SanzSernaProblem < otp.Problem
```
Steven-Roberts, line 1: Let's add a link to the paper where we got the problem.  We can also try to think of names other than SanzSerna.

src/+otp/+sanzserna/fforcing.m
```diff
@@ -0,0 +1,10 @@
+function dy = fforcing(t, y, ~, x)
+
+hIn = length(y);
```
AndreyAPopov, line 3: Minor point, but length could return the wrong thing here if the input is vectorized. size(y, 1) would probably be more correct here.

src/+otp/+sanzserna/SanzSernaProblem.m
```diff
@@ -4,6 +4,11 @@
     % applied to evolutionary problems in partial differential equations. Numerische Mathematik, 50(4), 405–418.
     % https://doi.org/10.1007/BF01396661
     
+    properties (SetAccess = private)
+        RhsLinear
+        RhsForcing
```
AndreyAPopov, line 9: Wouldn't it be better to somehow group the various splittings into one? I know we had this discussion before, and we did not come to a good conclusion. I personally would prefer something like

RhsLinearForcing, where RhsLinearForcing{1} is RhsLinear, and RhsLinearForcing{2} is RhsForcing, but...

src/+otp/+sanzserna/SanzSernaProblem.m
```diff
@@ -21,6 +26,11 @@ function onSettingsChanged(obj)
             obj.Rhs = otp.Rhs(@(t, y) otp.sanzserna.f(t, y, D, x), ...
                 otp.Rhs.FieldNames.Jacobian, @(t, y) otp.sanzserna.jac(t, y, D, x), ...
                 otp.Rhs.FieldNames.JacobianVectorProduct, @(t, y, v) otp.sanzserna.jvp(t, y, v, D, x));
+            
+            obj.RhsLinear = otp.Rhs(@(t, y) otp.sanzserna.flinear(t, y, D, x), ...
+                otp.Rhs.FieldNames.Jacobian, @(t, y) otp.sanzserna.jaclinear(t, y, D, x));
+            obj.RhsForcing = otp.Rhs(@(t, y) otp.sanzserna.fforcing(t, y, D, x), ...
+                otp.Rhs.FieldNames.Jacobian, @(t, y) otp.sanzserna.jacforcing(t, y, D, x));
```
AndreyAPopov, line 33: as we have jacobian vector products on the Rhs, we might as well do it on these as well.

Steven-Roberts: I can't imagine a scenario where we would need it and don't think it's worth adding.

src/+otp/+sanzserna/fforcing.m
Steven-Roberts: True.  This probably appears in many other problems as well.  It shouldn't matter unless vectorization is set to 'on' in the Rhs, but is still worth fixing.

PR #15, opened 2021-05-02
Lotka–Volterra Equations

AndreyAPopov:
Also know as predator-prey.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ ODE Test Problems can be installed by running the MATLAB script `install` from t
 
 ```matlab
 % Create a problem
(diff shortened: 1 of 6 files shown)


Steven-Roberts:
Looks ready to merge

AndreyAPopov:
Should we add more presets before merging?

Steven-Roberts:
I don't think so.  Different parameters will still lead to oscillatory dynamics.

AndreyAPopov:
Added a "stable" preset which has the initial condition on the attractor, thus constant for all time.

PR #16, opened 2021-05-07
QG Rhs optimization

AndreyAPopov:
Take the kron form of the equations and do everything in 2d, and also solve the sylvester equation with an eigenvalue approach.

TODO: solve the sylvester equation through discrete sine transforms.


src/+otp/+qg/QuasiGeostrophicProblem.m
```diff
@@ -166,7 +176,7 @@ function onSettingsChanged(obj)
             F = sin(pi*(ymat.' - 1));
             
             obj.Rhs = otp.Rhs(@(t, psi) ...
-                otp.qg.f(psi, Lx, Ly, P1, P1T, P2, P2T, L12, Dx, Dy, F, Re, Ro), ...
+                otp.qg.f(psi, Lx, Ly, P1, P1T, P2, P2T, L12, Dx, DyT, F, Re, Ro), ...
```
Steven-Roberts, line 179: Is it really necessary to pass in P1T and P2T?  It is more storage and arguments and I wouldn't think it's more efficient than transposing in f.

AndreyAPopov: I've timed it, and it is in fact faster to pre-compute the transposes. I am going to replace this with fast sine transforms once I figure out the correct way to do that anyways, so this will not exist in the final pull request (this is a draft).

AndreyAPopov: Actually in timings, this is faster than doing an fft. I assume that has a lot of overhead, so I will keep this version for now.

src/+otp/+qg/QuasiGeostrophicProblem.m
```diff
@@ -13,7 +13,6 @@
         DistanceFunction
         FlowVelocityMagnitude
         JacobianFlowVelocityMagnitudeVectorProduct
```
Steven-Roberts, line 15: These should have SetAccess private or protected

PR #18, opened 2021-05-17
Preset for optional symbolic integration of VanderPol problem 

elswit:


src/+otp/+vanderpol/+presets/ExtendedPrecision.m
```diff
@@ -0,0 +1,28 @@
+classdef ExtendedPrecision < otp.vanderpol.VanderpolProblem
+    methods
+        function obj = ExtendedPrecision(epsilon)
+            if nargin < 1
+                epsilon = sym(10^-6);
```
AndreyAPopov, line 5: maybe use "1e-6"? I would think that is more standard.

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
```diff
@@ -0,0 +1,28 @@
...
+                              -7300255173432526840/617673396283947, ...
+                              39058676403918956/22876792454961, ...
+                              -231923508930412/847288609443, ...
+                              515714746118/10460353203, ...
+                              -3927188888/387420489, ...
+                              34896076/14348907,...
+                              -1121308/1594323, ...
+                              15266/59049, ...
+                              -292/2187, 10/81, -2/3].';
+            y0 = [2; z0];
```
AndreyAPopov, line 19: Would we want to use sym(2)?

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
```diff
@@ -0,0 +1,28 @@
...
+            z0 = (epsilon.^(11:-1:0))*[1499571590693390159668/16677181699666569, ...
+                              -7300255173432526840/617673396283947, ...
+                              39058676403918956/22876792454961, ...
+                              -231923508930412/847288609443, ...
+                              515714746118/10460353203, ...
+                              -3927188888/387420489, ...
+                              34896076/14348907,...
+                              -1121308/1594323, ...
+                              15266/59049, ...
+                              -292/2187, 10/81, -2/3].';
```
AndreyAPopov, line 18: does this first convert "1499571590693390159668/16677181699666569" to double, then do the multiplication by sym?

Steven-Roberts: Yes, there is loss of precision here

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
AndreyAPopov: actually I think we would want to do "sym(10)^(-6)" to not lose any precision.

src/+otp/+vanderpol/VanderpolProblem.m
```diff
@@ -27,8 +27,8 @@ function onSettingsChanged(obj)
         function validateNewState(obj, newTimeSpan, newY0, newParameters)
             validateNewState@otp.Problem(obj, newTimeSpan, newY0, newParameters)
             
-            otp.utils.StructParser(newParameters) ...
-                .checkField('epsilon', 'scalar', 'real', 'finite', 'nonnegative');
+             otp.utils.StructParser(newParameters) ...
+                 .checkField('epsilon', 'scalar', 'real', 'finite', 'nonnegative');
```
Steven-Roberts, line 31: Make sure spacing is consistent.  Matlab's formatter (Ctrl-I) can help

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
Steven-Roberts: I think `vpa(1e-6)` is what we want.

Steven-Roberts:
Symbolics and vpa seems to break several things:

- StructParsers including the 'numeric' option
- Validating tspan and y0
- Anything else?

src/+otp/Problem.m
```diff
@@ -120,12 +120,12 @@ function validateNewState(obj, newTimeSpan, newY0, newParameters)
             % Ensures the TimeSpan, Y0, and Parameters are valid
             if length(newTimeSpan) ~= 2
                 error('TimeSpan must be a vector of two times');
-            elseif ~isnumeric(newTimeSpan)
-                error('TimeSpan must be numeric');
+            elseif ~( isnumeric(newTimeSpan) || all(isSymType(newTimeSpan,'number')))
+                error('TimeSpan must be numeric or symbolic ');
```
Steven-Roberts, line 124: Remove "or symbolic"

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
elswit: Doesn't matter in this case. 
sym(10^-6) - vpa(1e-6) = 0.0

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
elswit: Here, it does matter. The roundoff is actually large ~ 1e-12

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
Steven-Roberts: It does matter. Vpa has finite precision and behaves differently than sym.  I don't think sym is what we want

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
```diff
@@ -0,0 +1,29 @@
+classdef ExtendedPrecision < otp.vanderpol.VanderpolProblem
+    methods
+        function obj = ExtendedPrecision(epsilon)
+            if nargin < 1
+                epsilon = sym(10^-6);
+            end
+            
+            tspan = ([0, 0.5]);
+            z0 = (epsilon.^(11:-1:0))*[sym(1499571590693390159668)/16677181699666569, ...
```
Steven-Roberts, line 9: This is still converted to double first.  You need to use use string input like `sym('1499571590693390159668/16677181699666569')` to avoid losing digits.

AndreyAPopov: In fact, if we want to actually use vpa, then we would want to define the precision, then do this

```
vpa(sym('1499571590693390159668/16677181699666569')/sym('16677181699666569'), precision)
```

where `precision` is the number of digits we want to keep (have it be an input to the preset?)

Steven-Roberts: It's simpler to do `vpa('1499571590693390159668/16677181699666569')`

src/+otp/Problem.m
```diff
@@ -120,12 +120,12 @@ function validateNewState(obj, newTimeSpan, newY0, newParameters)
...
-            elseif ~isnumeric(newTimeSpan)
-                error('TimeSpan must be numeric');
+            elseif ~(isnumeric(newTimeSpan) || all(isSymType(newY0,'number')))
+                error('TimeSpan must be numeric or symbolic number');
             elseif ~iscolumn(newY0)
                 error('Y0 must be a column vector');
-            elseif ~(isnumeric(newY0) && all(isfinite(newY0)))
-                error('Y0 must be numeric and finite');
+            elseif ~((isnumeric(newY0) || all(isSymType(newY0,'number')) ) && all(isfinite(newY0)))
+                error('Y0 must be a finite numeric or symbolic number');
```
Steven-Roberts, line 128: Again I see no need to include "or symbolic".  We really don't need to support all symbolic, just vpa.

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
AndreyAPopov: from my understanding, chcracter vector calls to vpa are discouraged. I think rationals are fine though. hmmm

src/+otp/Problem.m
AndreyAPopov: Also, `isSymType` is part of the symbolic toolbox, thus having this function be part of `Problem.m` is a bad idea. I think I have a trick to do this that works with vpa without actually needing the symbolic toolbox:

```
function l = isnumericBsym(x)

try 
    x = double(x);
    l = all(isnumeric(x));
catch
    l = false;
end

end
```

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
elswit: So right now, if the user provides epsilon as  ``sym(1)/100``  the initial condition would be symbolic rational numbers. If it's set is vpa(1e-2, 30) then Y0 is type symbolic with 30 digits. If it's double output is double too. I think that's the best that can be done.  Question is what should be the default case. I vote for symbolic rational numbers, cuz the user can always truncate them later.

src/+otp/+vanderpol/+presets/ExtendedPrecision.m
```diff
@@ -0,0 +1,37 @@
...
+    methods
+        function obj = ExtendedPrecision(epsilon)
+            if nargin < 1
+                epsilon = 1e-6;
+            end
+            
+            cls = class(epsilon);
+            
+            coeffs = [ ...
+                cast(199944165973, cls) / cast(3486784401, cls) * cast(7499951716, cls) / cast(4782969, cls), ...
```
Steven-Roberts, line 11: I had to express `1532057748884056/16677181699666569` in terms of numbers that fit into a double.  Perhaps there is a more concise way to do this.

PR #22, opened 2021-06-30
Class-property-based validation instead of validateNewState and StructParser

Steven-Roberts:
I have only updated the Problem class and Brusselator so far.  After a review of that, I'll update the other problems.

### DIFF
--- a/src/+otp/+brusselator/BrusselatorParameters.m
+++ b/src/+otp/+brusselator/BrusselatorParameters.m
@@ -1,0 +1,6 @@
+classdef BrusselatorParameters
+    properties
+        ReactionRateA (1, 1) {mustBeReal, mustBePositive, mustBeFinite} = 1
+        ReactionRateB (1, 1) {mustBeReal, mustBePositive, mustBeFinite} = 1
+    end
+end
--- a/src/+otp/+brusselator/BrusselatorParameters.m
+++ b/src/+otp/+brusselator/BrusselatorParameters.m
@@ -1,6 +1,6 @@
 classdef BrusselatorParameters
     properties
-        ReactionRateA (1, 1) {mustBeReal, mustBePositive, mustBeFinite} = 1
-        ReactionRateB (1, 1) {mustBeReal, mustBePositive, mustBeFinite} = 1
+        ReactionRateA(1, 1) {mustBeReal, mustBePositive, mustBeFinite} = 1
+        ReactionRateB(1, 1) {mustBeReal, mustBePositive, mustBeFinite} = 1
     end
 end
--- a/src/+otp/+utils/+validation/mustBeNumericOrSym.m
+++ b/src/+otp/+utils/+validation/mustBeNumericOrSym.m
@@ -1,0 +1,7 @@
+function mustBeNumericOrSym(x)
+
+if ~isnumeric(x) && ~isa(x, 'sym')
+    throwAsCaller(MException('mustBeNumericOrSym:notNumericOrSym', 'Value must be numeric or symbolic.'));
+end
+
+end
--- a/src/+otp/Problem.m
+++ b/src/+otp/Problem.m
@@ -1,19 +1,15 @@
 classdef (Abstract) Problem < handle
     
     properties (SetAccess = immutable)
-        Name % A human-readable representation of the problem
-    end
-    
-    properties (SetAccess = immutable, GetAccess = private)
-        ExpectedNumVars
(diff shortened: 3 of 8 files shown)


src/+otp/Problem.m
```diff
@@ -116,21 +111,8 @@
     end
     
     methods (Access = protected)
-        function validateNewState(obj, newTimeSpan, newY0, newParameters)
+        function validateNewState(~, ~, ~, ~)
```
AndreyAPopov, line 114: I assume we don't actually need this function now?

src/+otp/ProblemSettings.m
```diff
@@ -0,0 +1,18 @@
+classdef (Hidden) ProblemSettings
+    properties
+        ExpectedNumVars(1, 1) double {mustBeNonnegative}
```
AndreyAPopov, line 3: The spacing here is inconsistent with what you do in the brusselator one. Pick one IMO.


Also: why "double"?

src/+otp/+brusselator/BrusselatorParameters.m
```diff
@@ -0,0 +1,6 @@
+classdef BrusselatorParameters
+    properties
+        ReactionRateA (1, 1) {mustBeReal, mustBePositive, mustBeFinite} = 1
```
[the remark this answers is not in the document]
Steven-Roberts: https://www.mathworks.com/help/matlab/matlab_oop/validate-property-values.html

src/+otp/Problem.m
Steven-Roberts: Subclasses might need to override it.  For example pendulum can show warning if number of masses is different than number of bob lengths.

src/+otp/+brusselator/BrusselatorParameters.m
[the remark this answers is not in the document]
AndreyAPopov: Thanks

src/+otp/Problem.m
AndreyAPopov: Hmm, I agree, but, then should it be abstract? Would that break anything?

src/+otp/Problem.m
```diff
@@ -27,47 +23,46 @@
...
         
         function parameters = get.Parameters(obj)
-            parameters = obj.Settings.parameters;
+            parameters = obj.Settings.Parameters;
         end
         
         function set.Settings(obj, value)
-            obj.validateNewState(value.timeSpan, value.y0, value.parameters);
+            expNumVars = value.ExpectedNumVars;
+            if ~(isempty(expNumVars) || length(value.Y0) == expNumVars)
```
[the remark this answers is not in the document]
Steven-Roberts: If I change it here, there are dozens of other places where it would need to change.  `Y0` is guaranteed to be a vector, so they are equivalent.  Most people are more familiar with `length`.

src/+otp/Problem.m
Steven-Roberts: The vast majority of problems don't need to do validation (past what `Problem` already does).  Adding abstract would make these problems implement empty functions.

AndreyAPopov: Makes sense.

src/+otp/+utils/+validation/mustBeNumericOrSym.m
```diff
@@ -0,0 +1,7 @@
+function mustBeNumericOrSym(x)
+
+if ~isnumeric(x) && ~isa(x, 'sym')
+    throwAsCaller(MException('mustBeNumericOrSym:notNumericOrSym', 'Value must be numeric or symbolic.'));
```
AndreyAPopov, line 4: oh, I like this. Very  fancy!

src/+otp/Problem.m
```diff
@@ -53,7 +53,7 @@
         
         function set.Settings(obj, value)
             expNumVars = value.ExpectedNumVars;
-            if ~(isempty(expNumVars) || length(value.Y0) == expNumVars)
+            if expNumVars > 0 && length(value.Y0) ~= expNumVars
```
AndreyAPopov, line 56: I would still like expNumVars to be possibly empty, that is very standard in matlab IMO

AndreyAPopov: Not a strong opinion on it though.

Steven-Roberts:
Closing because superseded by #24

PR #24, opened 2021-10-30
Support for Octave

Steven-Roberts:
Some things that have changed/broken

- Abstract methods are not supported.  I have emulated them with functions that throw an error if not overridden.  Not ideal but doesn't break subclasses.
- Immutable is not supported.  Effectively immutable properties are now just private.
- Movies that use `animatedline`, which is most, are not supported.  Not worried about supporting that.  I manged to get `LineMovie` working at least.
- Parameters are classes instead of a struct.  This was planned anyway, but class property validation is not supported in Octave.  Could have the validation syntax in comments then use tool to remove comments to export version of code for Matlab.
- Rhs has been modified because `dynamicprops` are not supported.  Rhs properties are now fixed.  A couple problems that use custom Rhs properties will need reevaluation.
- Octave does not have `odextend` and the format of solution structures for event problems appears different.


src/+otp/+arenstorf/+presets/Canonical.m
```diff
@@ -1,17 +1,36 @@
...
+            %
+            %   OBJ = CANONICAL() Uses a MU corresponding to the Moon
+    
+            if nargin < 1
+                mu = 0.012277471;
+            end
             
-            params.m1 = s.m1;
-            params.m2 = 1-params.m1;
+            params = otp.arenstorf.ArenstorfParameters;
```
AndreyAPopov, line 23: Does octave not support input parser?

src/+otp/+arenstorf/ArenstorfParameters.m
```diff
@@ -0,0 +1,10 @@
+classdef ArenstorfParameters
+    %ARENSTORFPARAMETERS User-configurable parameters for the Arenstorf problem
+    %
+    %   See also otp.arenstorf.ArenstorfProblem
+    
+    properties
+        %MU The mass of one body, while the other body has mass 1-Mu
+        Mu %MATLAB ONLY: (1, 1) {mustBeReal, mustBeFinite}
```
AndreyAPopov, line 8: Does "mustBeReal" work with symbolic? it looks like you are trying to support it with the classes, but I am not sure that the matlab code here will work with it.

src/+otp/+robertson/RobertsonParameters.m
```diff
@@ -0,0 +1,7 @@
+classdef RobertsonParameters
+    properties
+        K1
```
AndreyAPopov, line 3: These are all reaction coefficients, so they are also 

```%MATLAB ONLY: (1,1) {..........}```` finite, and positive

File: src/+otp/+utils/+movie/+recorder/VideoRecorder.m [hunk 4]
```diff
@@ -1,11 +1,28 @@
...
-    methods (Abstract)
-        frameRate = getFrameRate(obj);
-        setFrameRate(obj, newFrameRate);
-        start(obj, totalFrames);
-        recordFrame(obj, fig);
-        stop(obj);
-        h = play(obj);
+    methods
+        function frameRate = getFrameRate(obj)
+            error('Abstract method getFrameRate must be implemented by a subclass');
```
AndreyAPopov, line 4: I assume this is an octave workaround. That's fine.

src/+otp/+utils/FancyPlot.m
```diff
@@ -53,11 +53,11 @@ function axisLimits(ax, dir, data, padding)
             
             switch get(ax, strcat(dir, 'scale'))
                 case 'linear'
-                    [yMin, yMax] = bounds(data, 'all');
```
AndreyAPopov, line 56: octave does not support 'all'?

src/+otp/Problem.m
```diff
@@ -116,6 +112,11 @@
     end
     
     methods (Access = protected)
+        % This method is called when either TimeSpan, Y0, or parameters are changed.  It should update F and other properties such as a Jacobian to reflect the changes.  This is effevtively an abstract function but not explicitly marked so in order to support Octave
+        function onSettingsChanged(obj)
+            error('Abstract method onSettingsChnaged must be implemented by a subclass');
```
AndreyAPopov, line 117: typo

src/+otp/Problem.m
```diff
@@ -225,6 +225,7 @@ function validateNewState(obj, newTimeSpan, newY0, newParameters)
                 end
                 
                 options = problem.Rhs.odeset(options);
+                % TODO: Octave does not support odextend
                 sol = odextend(sol, problem.Rhs.F, problem.TimeSpan(end), problem.Y0, options);
```
AndreyAPopov, line 229: odextend basically just merges the structures. I think we can have octave specific code here.

src/+otp/+arenstorf/ArenstorfParameters.m
Steven-Roberts: `mustBeReal` works with symbolic and vpa in matlab but not octave.  That seems like an octave bug, so I say we keep the validation as is.

src/+otp/+arenstorf/+presets/Canonical.m
Steven-Roberts: It does, but it seemed unnecessary for a one-input constructor.

src/+otp/+utils/+movie/+recorder/VideoRecorder.m
Steven-Roberts: Yes.  This emulates an abstract method.

src/+otp/+utils/FancyPlot.m
Steven-Roberts: Unfortunately, no

src/+otp/Problem.m
Steven-Roberts: We could implement this, but it might be hard to maintain.  If the solution struct changes in a future release we would need to be able to support multiple formats.

src/+otp/+arenstorf/+presets/Canonical.m
AndreyAPopov: Sure

ODE Test Problems.prj
```diff
@@ -1,90 +1,90 @@
...
-    </fileset.rootfiles>
-    <fileset.depfun.included />
-    <fileset.depfun.excluded />
-    <fileset.package />
-  </configuration>
-</deployment-project>
+<deployment-project plugin="plugin.toolbox" plugin-version="1.0">
+  <configuration name="ODE Test Problems" target="target.toolbox" target-name="Package Toolbox">
+    <param.appname>ODE Test Problems</param.appname>
+    <param.authnamewatermark>Steven Roberts, Andrey Popov, Ardian Sandu</param.authnamewatermark>
```
AndreyAPopov, line 4: Add arash?

src/+otp/Problem.m
AndreyAPopov: I think solution structure will not change, as it is a public API. Matlab would rather depricate all the "odeXX" functions and make new ones, than change their output.

src/+otp/+arenstorf/+presets/Canonical.m
```diff
@@ -1,39 +1,40 @@
...
+                cast(0, cls); ...
+                cast(-823970832321143, cls) / cast(411659154384760, cls)];
+            tspan = [cast(0, cls); ...
+                cast(4541277234950502, cls) / cast(266113073862361, cls)];
+            
+            obj = obj@otp.arenstorf.ArenstorfProblem(tspan, y0, params);
+        end
+
+    end
 end
```
AndreyAPopov, line 40: newline

src/+otp/Problem.m
```diff
@@ -129,6 +129,8 @@ function validateNewState(obj, newTimeSpan, newY0, newParameters)
                 error('Y0 must be numeric');
             elseif ~(isempty(obj.ExpectedNumVars) || length(newY0) == obj.ExpectedNumVars)
                 error('Expected Y0 to have %d components but has %d', obj.ExpectedNumVars, length(newY0));
+            elseif isempty(newParameters)
+                error('Parameters cannot be empty');
```
AndreyAPopov, line 133: This is interesting. What if the problem has no parameters? Do we want to force an empty parameter class? I am fine either way.

src/+otp/Problem.m
Steven-Roberts: I'll give a shot to octave support.  I did notice octave transposed some things for event problems which might be annoying.

src/+otp/Problem.m
Steven-Roberts: Yes, it would require a class with no properties.  This makes documentation and usage more consistent.

AndreyAPopov:
I started the conversion of other test problems to the new parameters format. Lorenz 63 seems to work for now.

I also fixed the install script to work with matlab and windows (well windows does not do the replacing as-of-yet).

AndreyAPopov:
Is Transistor Amplifier the only problem left to update?

src/+otp/+sanzserna/SanzSernaProblem.m
```diff
@@ -38,7 +38,7 @@ function onSettingsChanged(obj)
         end
         
         function sol = internalSolve(obj, varargin)
-            sol = internalSolve@otp.Problem(obj, 'Method', @ode45, varargin{:});
+            sol = internalSolve@otp.Problem(obj, 'Solver', otp.utils.Solver.Nonstiff, varargin{:});
```
AndreyAPopov, line 41: should this be NonStiff?

src/+otp/+utils/+compatibility/isoctave.m
```diff
@@ -0,0 +1,15 @@
...
+    end
+
+    if nargin == 0
+        retVal = cacheval;
+    elseif cacheval
+        retVal = octaveVal;
+    else
+        retVal = matlabVal;
+    end
+end
```
AndreyAPopov, line 15: This needs a new line

src/+otptests/validateallpresets.m
```diff
@@ -2,42 +2,48 @@
...
-                        fprintf('---- The preset %s of the model %s has failed to solve\n', presetname, modelname)
-                    end
-                    fprintf('The preset %s of the model %s has been solved successfully\n', presetname, modelname);
-                end
+        try
+            if strcmp(modelname, 'qg')
+                model.TimeSpan = [0, 0.01];
+            end
+            if strcmp(modelname, 'e5')
+                model.TimeSpan = [0, 1e-9];
```
AndreyAPopov, line 38: should this be octave only?

src/+otp/+sanzserna/SanzSernaProblem.m
AndreyAPopov: as per discussion offline: could go either way

AndreyAPopov:
> I think we can always use w flag

done

PR #27, opened 2022-01-18
QG Approximate deconvolution large eddy simulation

AndreyAPopov:
Reworked resizing to be more general, and added AD LES with the new way of solving the Sylvester equation form of the poisson equation. Everything should hopefully be faster now. Do not look forward to doing the derivatives.


src/+otp/+qg/+presets/Canonical.m
```diff
@@ -8,19 +8,22 @@
...
             
-            [nx, ny] = otp.qg.QuasiGeostrophicProblem.name2size(s.Size);
+            nx = s.Size(1);
+            ny = s.Size(2);
             
             params = otp.qg.QuasiGeostrophicParameters;
             params.Nx = nx;
             params.Ny = ny;
             params.Re = s.ReynoldsNumber;
             params.Ro = s.RossbyNumber;
```
Steven-Roberts, line 24: The parameters and option names should be consistent

src/+otp/+qg/QuasiGeostrophicParameters.m
```diff
@@ -10,5 +10,9 @@
         Re %MATLAB ONLY: (1,1) {mustBeFinite, mustBePositive} = 450
         %Ro is the Rossby number
         Ro %MATLAB ONLY: (1,1) {mustBeFinite} = 0.0036
+        %ADlambda is the approximate deconvolution parameter
+        ADLambda %MATLAB ONLY: (1,1) {mustBeFinite, mustBePositive} = 1
```
Steven-Roberts, line 14: I'm not familiar with this approximate filter, but is there a set of params that allow it to be disabled?  Unless it is very commonly used, I don't want to force approximations to the underlying PDE

Steven-Roberts: Nevermind, it's in a separate RHS.  Didn't review far enough.  This might cause some confusion when params affect one RHS but not another.

src/+otp/+qg/f.m
```diff
@@ -2,31 +2,31 @@
 
 [nx, ny] = size(L12);
 
-psi = reshape(psi, nx, ny);
+psi = reshape(psi, nx, ny, []);
 
 % Calculate the vorticity
-q = -(Lx*psi + psi*Ly);
+q = -(pagemtimes(Lx, psi) + pagemtimes(psi, Ly));
```
Steven-Roberts, line 8: Unfortunately, `pagemtimes` is not supported in Octave.  I'm a bit more concerned about the lack or sparse support.  Is this only used to support vectorized psi?

AndreyAPopov: Lack of sparse support is due to pagemtimes being built for dlarray. That will go away in the future we assume.

It is used for vectorizing the RHS, but also vectorizing jacobian and adjoint vector products, so it is pretty important. We can add an octave-only stub function to have naive for-loop support, if needed.

src/+otp/+qg/QuasiGeostrophicParameters.m
AndreyAPopov: >This might cause some confusion when params affect one RHS but not another.

I'm not sure of a good way to solve this issue. This will not be the last problem that has weird RHS issues I assume, so finding a good compromise right now is key IMO.

AndreyAPopov:
I know that this uses things not currently supported by octave, but we are using this right now, and very actively. Unless you have any other objections, I want to merge @Steven-Roberts

Steven-Roberts:
Ok, let's merge it then.  I'll create an issue so we don't forget the Octave incompatibility.

PR #33, opened 2022-02-01
Operator overloading on RHS

AndreyAPopov:
This is a test as of now.

### DIFF
--- a/src/+otp/RHS.m
+++ b/src/+otp/RHS.m
@@ -1,4 +1,4 @@
-classdef RHS   
+classdef RHS < matlab.mixin.indexing.RedefinesParen
     properties (SetAccess = private)
         F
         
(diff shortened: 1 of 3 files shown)


AndreyAPopov:
So here is how all this works in matlab:

```
>> model = otp.brusselator.presets.Canonical;
>> firstCompRHS = model.RHS(1, :);
>> secondCompRHS = model.RHS(2, :);
>> reverseRHS = [secondCompRHS; firstCompRHS];
>> model.RHS.F(0, model.Y0)
ans =
    -6
     5
>> reverseRHS.F(0, model.Y0)
ans =
     5
    -6
```

Steven-Roberts:
This should be ready to merge after some testing

src/+otp/RHS.m
```diff
@@ -48,5 +118,99 @@
...
+            
+            % JPattern is problematic to compute for division operators due to
+            % singular patterns
+            
+            % Mass matrices introduce several difficulties. When singular, it
+            % makes it infeasible to update InitialSlope, and therefore, it is
+            % always unset. To avoid issues with two RHS' having different mass
+            % matrices, only the primary RHS is used.
+            
+            newRHS = otp.RHS(f, ...
```
AndreyAPopov, line 180: Mass matrices are a big difficulty right now it seems like.

AndreyAPopov:
I feel like what we need now is to setup some pretty serious edge-case tests to exhaustively test every possible scenario

AndreyAPopov:
As it is now, there is effectively no guarantees that this will work as intended

src/+otp/RHS.m
Steven-Roberts: I don't think there is much more that can be done outside of warnings for incompatible mass matrices.

AndreyAPopov: I agree. That is fine

Steven-Roberts:
I say we go ahead and merge this.  Clearly it needs additional testing, but proper testing for OTP will be a large undertaking in a separate PR.

AndreyAPopov:
> I say we go ahead and merge this. Clearly it needs additional testing, but proper testing for OTP will be a large undertaking in a separate PR.

Fine. It is in a usable state. We will get more mileage out of testing by people actually using it.

Issue #30, opened 2022-01-22
Validation for problems where size of parameter is related to NumVars


Steven-Roberts:
It might be helpful on problems like `NBody` and `Pendulum` to display a warning when the number of masses differs from `NumVars`.  This would occur in `validateNewState`.

PR #34, opened 2022-03-26
Add warnings when parameters are inconsistent with length of Y0

Steven-Roberts:
Addresses #30


AndreyAPopov:
I really like using warnID, so that warnings of a certain type can be disabled.

I suggest 'OTP:inconsistentInitialConditionSize'

Steven-Roberts:
Ready to merge if you are @AndreyAPopov

AndreyAPopov:
Looks good.

PR #36, opened 2022-05-10
Constrained pendulum problem

AndreyAPopov:
For use in half-explicit methods and for data assimilation.

### DIFF
--- a/src/+otp/+utils/+solvers/dae43.m
+++ b/src/+otp/+utils/+solvers/dae43.m
@@ -6,7 +6,7 @@ h = odeget(options, 'InitialStep', []);
 reltol = odeget(options, 'RelTol', 1e-3);
 abstol = odeget(options, 'AbsTol', 1e-6);
 J = odeget(options, 'Jacobian', []);
-M = odeget(options, 'Mass', speye(numel(y0)));
+M = odeget(options, 'Mass', []);
 MStateDependence = odeget(options, 'MStateDependence', 'none');
 
 % We won't support state-dependent Mass, simple as that
@@ -15,7 +15,7 @@ if strcmp(MStateDependence, 'strong')
 end
 
 if isempty(J)
-    J = @(t, y) jacapprox(f, t, y);
+    J = @(t, y) otp.utils.jacobianApproximation(f, t, y);
     usesparse = false;
 else
     usesparse = issparse(J(tspan(1), y0));
@@ -29,6 +29,8 @@ if isempty(M)
     end
 elseif ~isa(M, 'function_handle')
     M = @(t, y) M;
+elseif isa(M, 'function_handle') && nargin(M) == 1
+    M = @(t, y) M(t);
 end
 
 % Lobatto IIIC
@@ -42,7 +44,7 @@ orderE = 3;
 
 % Compute initial step size
 if isempty(h)
-    sc = abstol + reltol*abs(y0);
+    sc = abstol + reltol.*abs(y0);
     f0 = f(tspan(1), y0);
     d0 = sqrt(mean((y0./sc).^2));
     d1 = sqrt(mean((f0./sc).^2));
@@ -69,10 +71,15 @@ tend = tspan(end);
 
 step = 1;
 
-Mc = M(tspan(1), y0);
-Mfull = zeros(n*stagenum, n*stagenum, 'like', Mc);
+if usesparse
+    Mfull = sparse(n*stagenum, n*stagenum);
+    Jfull = sparse(n*stagenum, n*stagenum);
+else
+    Mfull = zeros(n*stagenum, n*stagenum);
+    Jfull = zeros(n*stagenum, n*stagenum);
+end
+
 gfull = zeros(n*stagenum, 1);
-Jfull = zeros(n*stagenum, n*stagenum, 'like', Mc);
 newtonk = zeros(n*stagenum, 1);
 sc = zeros(n*stagenum, 1);
 
@@ -90,7 +97,7 @@ while tc < tend
         Mfull(si, si) = M(tc + h*c(stage), yc);
     end
 
-    ntol = min(abstol, reltol);
+    ntol = min(abstol, min(reltol));
     nmaxits = 1e3;
     its = 0;
     etak = inf;
@@ -124,7 +131,7 @@ while tc < tend
             si = ((stage - 1)*n + 1):(stage*n);
             staget = tc + c(stage)*h;
             ycs = yc + h*reshape(newtonk, n, [])*(A(stage, :).');
-            sc(si) = abstol + reltol*abs(ycs);
+            sc(si) = abstol + reltol.*abs(ycs);
             Mc = Mfull(si, si);
             gfull(si) =  Mc*newtonk(si) - f(staget, ycs);
         end
@@ -191,7 +198,7 @@ while tc < tend
     yhat = yc + h*reshape(newtonk, n, [])*bhat.';
     ycnew = yc + h*reshape(newtonk, n, [])*b.';
 
-    sc = abstol + max(abs(ycnew), abs(yc))*reltol;
+    sc = abstol + max(abs(ycnew), abs(yc)).*reltol;
 
     Mc = M(tc + h , ycnew);
 
@@ -227,23 +234,6 @@ if nargout < 2
     sol = struct('x', t, 'y', y.');
 else
     sol = t;
-end
-
-end
-
-function J = jacapprox(f, t, y)
-
-n = numel(y);
-
-J = zeros(n, n);
-
-h = sqrt(1e-6);
-
-f0 = f(t, y);
-
-for i = 1:n
-    e = zeros(n, 1); e(i) = 1;
-    J(:, i) = (f(t, y + h*e) - f0)/h;
 end
 
 end
--- a/src/+otp/+utils/jacobianApproximation.m
+++ b/src/+otp/+utils/jacobianApproximation.m
@@ -1,0 +1,17 @@
+function J = jacobianApproximation(f, t, y)
+% TODO: make this more robust
+
+n = numel(y);
+
+J = zeros(n, n);
+
+h = sqrt(1e-6);
+
+f0 = f(t, y);
+
+for i = 1:n
+    e = zeros(n, 1); e(i) = 1;
+    J(:, i) = (f(t, y + h*e) - f0)/h;
+end
+
+end


src/+otp/+constrainedpendulum/+presets/Canonical.m
```diff
@@ -0,0 +1,22 @@
+classdef Canonical < otp.constrainedpendulum.ConstrainedPendulumProblem
+    %CANONICAL The constrained pendulum problem
+    %
+    % See
+    %    Ascher, Uri. "On symmetric schemes and differential-algebraic equations."
+    %    SIAM journal on scientific and statistical computing 10.5 (1989): 937-949.
```
Steven-Roberts, line 6: Is this the right source?  I don't see anything on a pendulum.

src/+otp/+constrainedpendulum/ConstrainedPendulumParameters.m
```diff
@@ -0,0 +1,11 @@
+classdef ConstrainedPendulumParameters
+    %CONSTRAINEDPENDULUMPARAMETERS
+    properties
+        %Gravity defines the magnitude of acceleration towards the ground
+        Gravity %MATLAB ONLY: (1,1) {mustBeNonnegative}
+        %Mass defines the mass of the pendulum
+        Mass %MATLAB ONLY: (1,1) {mustBeNonnegative}
```
Steven-Roberts, line 7: Sharper validation: `{mustBeReal, mustBeFinite, mustBePositive}`

src/+otp/+constrainedpendulum/ConstrainedPendulumProblem.m
```diff
@@ -0,0 +1,35 @@
+classdef ConstrainedPendulumProblem < otp.Problem
```
Steven-Roberts, line 1: Perhaps just `PendulumDAEProblem`.

src/+otp/+constrainedpendulum/ConstrainedPendulumProblem.m
```diff
@@ -0,0 +1,35 @@
...
+    %CONSTRAINEDPENDULUM PROBLEM This is a Hessenberg Index-2 DAE problem posed in terms of three constraints
+    %
+    
+    properties
```
Steven-Roberts, line 5: These should have `SetAccess = private` or `protected`

src/+otp/+constrainedpendulum/ConstrainedPendulumParameters.m
Steven-Roberts: Better yet, it should match https://github.com/ComputationalScienceLaboratory/ODE-Test-Problems/blob/3d4cb2262931eaf261ca3061479b81928b1ef419/src/+otp/+pendulum/PendulumParameters.m

src/+otp/+constrainedpendulum/ConstrainedPendulumProblem.m
```diff
@@ -0,0 +1,35 @@
...
+        function onSettingsChanged(obj)
+            m = obj.Parameters.Mass;
+            l = obj.Parameters.Length;
+            g = obj.Parameters.Gravity;
+
+            % get initial energy
+            initialconstraints = otp.constrainedpendulum.constraints([], obj.Y0, g, m, l, 0);
+            E0 = initialconstraints(3);
+            
+            obj.RHS = otp.RHS(@(t, y) otp.constrainedpendulum.f(t, y, g, m, l, E0));
```
Steven-Roberts, line 27: This RHS will lead to confusion as a call to `solve` will not enforce constraints, right?  Ideally, `RHS` should be in a form that a built-in solver can integrate, but we can let that slide for an index-2 DAE.  I suggest the default RHS should use a mass matrix and an `f` that operates on the full state (diff and alg).  Then there can be a `RHSDifferential` and `RHSAlgebraic` to get the individual pieces.

Steven-Roberts: Also it looks like `'Vectorized', 'on'` can be added to RHS

src/+otp/+constrainedpendulum/constraintsjacobian.m
```diff
@@ -0,0 +1,27 @@
+function dc = constraintsjacobian(~, state, g, m, ~, ~)
```
Steven-Roberts, line 1: What Jacobian is this exactly?  For the DAE
y' = f(y, z)
0 = g(z)
is it g_z or f_z*g_z

src/+otp/+constrainedpendulum/ConstrainedPendulumProblem.m
Steven-Roberts: This class ought to override `internalIndex2label`

Steven-Roberts:
Long term, I would like to have a problem class for constrained mechanical systems (Solving ODEs II pg. 464), and the pendulum would simply be a preset.  This would also unify the handling of reduced index forms.

src/+otp/+constrainedpendulum/+presets/Canonical.m
amitns122: Hairer, E., Roche, M., Lubich, C. (1989). Description of differential-algebraic problems. In: The Numerical Solution of Differential-Algebraic Systems by Runge-Kutta Methods. Lecture Notes in Mathematics, vol 1409. Springer, Berlin, Heidelberg. https://doi.org/10.1007/BFb0093948

That is the pendulum problem source. 

The Ascher book has a nice formulation we use.

src/+otp/+constrainedpendulum/constraintsjacobian.m
amitns122: in this formulation. y' = f(y) - (g_y)^T z. T is transpose. 

so g_y*f_z = g_y*((g_y)^T)

src/+otp/+constrainedpendulum/+presets/Canonical.m
AndreyAPopov: Yeah the citation is wrong.

src/+otp/+constrainedpendulum/constraintsjacobian.m
AndreyAPopov: this jacobian is actually just g_y

src/+otp/+constrainedpendulum/ConstrainedPendulumProblem.m
AndreyAPopov: Agree with both

src/+otp/+constrainedpendulum/ConstrainedPendulumParameters.m
AndreyAPopov: Good idea

src/+otp/+constrainedpendulum/+presets/Canonical.m
```diff
@@ -0,0 +1,22 @@
...
+            tspan = [0; 10];
+            
+            params = otp.constrainedpendulum.ConstrainedPendulumParameters;
+            params.Mass = 1;
+            params.Length = 1;
+            params.Gravity = otp.utils.PhysicalConstants.EarthGravity;
+            
+            y0 = [sqrt(2)/2; sqrt(2)/2; 0; 0];
+            
+            obj = obj@otp.constrainedpendulum.ConstrainedPendulumProblem(tspan, y0, params);
```
reid-g: Will the `solve` functionality be disabled for this problem?

src/+otp/+constrainedpendulum/constraintsjacobian.m
reid-g: Is this to be the standard for otp when getting the Jacobian for Half-Explicit methods? Or is this an internal functionality of the problem?

src/+otp/+pendulumdae/fdifferential.m
```diff
@@ -0,0 +1,19 @@
+function dstate = fdifferential(~, state, g, m, ~, ~)
```
Steven-Roberts, line 1: The convention for functions is camel case now: https://github.com/ComputationalScienceLaboratory/ODE-Test-Problems/blob/master/docs/CONTRIBUTING.md#functions

src/+otp/+pendulumdae/PendulumDAEProblem.m
```diff
@@ -0,0 +1,54 @@
...
+            g = obj.Parameters.Gravity;
+
+            % get initial energy
+            initialconstraints = otp.pendulumdae.constraints([], obj.Y0Differential, g, m, l, 0);
+            E0 = initialconstraints(3);
+            
+
+            % The right hand size in terms of x, y, x', y', and three control parameters z
+            obj.RHS = otp.RHS(@(t, y) otp.pendulumdae.f(t, y, g, m, l, E0), ...
+                'Mass', @(t, y) otp.pendulumdae.mass(t, y, g, m, l, E0));
```
Steven-Roberts, line 34: Since the mass matrix is constant, it can be a matrix instead of a function handle.  See https://github.com/ComputationalScienceLaboratory/ODE-Test-Problems/blob/3d4cb2262931eaf261ca3061479b81928b1ef419/src/%2Botp/%2Bzlakinetics/ZlaKineticsProblem.m#L19 for example

src/+otp/+constrainedpendulum/+presets/Canonical.m
AndreyAPopov: That is an interesting question. In theory we should be able to add our own small method to OTP to handle these cases

src/+otp/+constrainedpendulum/constraintsjacobian.m
AndreyAPopov: if the problem is posed in terms of invariants, this should be standard. I might rename it to invariantJacobian

src/+otp/+pendulumdae/PendulumDAEProblem.m
AndreyAPopov: I was just copying the pendulum problem, which had @(t,y). I will change

AndreyAPopov:
Unless I am missing something, I think I have addressed all the issues.

src/+otp/+pendulumdae/PendulumDAEProblem.m
```diff
@@ -0,0 +1,87 @@
...
+    %
+    
+    properties (SetAccess=protected)
+        RHSDifferential
+        RHSAlgebraic
+    end
+
+    properties (Dependent)
+      Y0Differential
+      Z0Algebraic
```
Steven-Roberts, line 12: I'm not opposed to having these, but if we do include this, I think all DAE problems should have these for consistency.  Also, I think `Y0Algebraic` would be better than `Z0Algebraic`.

src/+otp/+pendulumdae/PendulumDAEProblem.m
```diff
@@ -0,0 +1,87 @@
...
+            obj.RHS = otp.RHS(@(t, y) otp.pendulumdae.f(t, y, g, m, l, E0), ...
+                'Mass', otp.pendulumdae.mass([], [], g, m, l, E0), ...
+                'JacobianVectorProduct', ...
+                @(t, y, v) otp.pendulumdae.jacobianVectorProduct(t, y, g, m, l, E0, v), ...
+                'Vectorized', 'on');
+
+            % Generate the constituent RHS for the differential part
+            obj.RHSDifferential = otp.RHS(@(t, y) otp.pendulumdae.fDifferential(t, y, g, m, l, E0), ...
+                'JacobianVectorProduct', ...
+                @(t, y, v) otp.pendulumdae.jacobianDifferentialVectorProduct(t, y, g, m, l, E0, v), ...
```
Steven-Roberts, line 41: `jacobianVectorProductDifferential` is the naming convention https://github.com/ComputationalScienceLaboratory/ODE-Test-Problems/blob/master/docs/CONTRIBUTING.md#functions-for-rhs

src/+otp/+pendulumdae/PendulumDAEProblem.m
```diff
@@ -0,0 +1,87 @@
...
+                'Vectorized', 'on');
+
+            % Generate the constituent RHS for the algebraic part
+            obj.RHSAlgebraic = otp.RHS(@(t, y) otp.pendulumdae.invariants(t, y, g, m, l, E0), ...
```
Steven-Roberts, line 45: The terms "algebraic" and "invariant" are used inconsistently.  I think "algebraic" is the more standard DAE term, so I say we should name the functions
`fAlgebraic`
`jacobianAlgebraic`
...

Steven-Roberts:
> Unless I am missing something, I think I have addressed all the issues.

Added some final pedantic comments

src/+otp/+constrainedpendulum/+presets/Canonical.m
Steven-Roberts: Good point.  Here's one idea: Add a `HighIndexDAE` property to https://github.com/ComputationalScienceLaboratory/ODE-Test-Problems/blob/master/src/%2Botp/%2Butils/Solver.m which simply throws an error message that high-index DAEs are not supported by matlab solvers.  Eventually when we have a constrained mechanical system problem, we will support RHS reduced to index-1 which can rely on built-in solvers.

src/+otp/+pendulumdae/PendulumDAEProblem.m
reid-g: Doesn't really differentiate the two. Algebraic implies there is no direct differential equation thus it is a variable obtained from differentiation of the constraint. I think it should stay.

Steven-Roberts: I think this naming is pretty clear.  It's mostly motivated by established naming conventions.  For partitioned problems, we use <property><partition name>.  E.g.

RHSStiff and RHSNonstiff
jacobianDifferential and jacobianAlgebraic
Y0Position and Y0Velocity

This is nice for autocompletion as well.

reid-g: My bad I misinterpreted what was being proposed. I thought you wanted `Y0Differential -> Y0Algebraic` not `Z0Algebraic -> Y0Algebraic`. I agree on the convention you proposed.

AndreyAPopov: I was thinking of changing it to that already, glad to see we all think alike

src/+otp/+pendulumdae/PendulumDAEProblem.m
```diff
@@ -0,0 +1,87 @@
...
+
+            % get initial energy
+            initialconstraints = otp.pendulumdae.invariants([], obj.Y0Differential, g, m, l, 0);
+            E0 = initialconstraints(3);
+
+            % The right hand size in terms of x, y, x', y', and three control parameters z
+            obj.RHS = otp.RHS(@(t, y) otp.pendulumdae.f(t, y, g, m, l, E0), ...
+                'Mass', otp.pendulumdae.mass([], [], g, m, l, E0), ...
+                'JacobianVectorProduct', ...
+                @(t, y, v) otp.pendulumdae.jacobianVectorProduct(t, y, g, m, l, E0, v), ...
```
Steven-Roberts, line 35: `v` should be the 3rd argument, and parameters should always be last.

AndreyAPopov:
Alright, The last think I want to add is the half-explicit jacobian, and Jvp, but I am not sure what to call it.

src/+otp/+utils/+solvers/dae34.m
```diff
@@ -0,0 +1,156 @@
+function [sol, y] = dae34(f, tspan, y0, options)
+% See
+%    Cameron, F., Palmroth, M., & Piché, R. (2002). 
+%    Quasi stage order conditions for SDIRK methods. 
+%    Applied numerical mathematics, 42(1-3), 61-75.
```
Steven-Roberts, line 5: The method in this paper is order 2 not order 3.  I think a method suitable for both index-2 and -3 DAEs would be preferable if we can find a good one.

src/+otp/+utils/+solvers/dae34.m
```diff
@@ -0,0 +1,156 @@
...
+
+f1 = f(tspan(1), y0);
+hdefault = norm(y0)/norm(f1)*0.01;
```
Steven-Roberts, line 8: Upgrade to procedure from Solving ODEs I pg 169

src/+otp/+utils/+solvers/dae34.m
```diff
@@ -0,0 +1,156 @@
...
+
+h = odeget(options, 'InitialStep', hdefault);
+reltol = odeget(options, 'RelTol', 1e-3);
+abstol = odeget(options, 'AbsTol', 1e-6);
+J = odeget(options, 'Jacobian', @(t, y) jacapprox(f, t, y));
+M = odeget(options, 'Mass', speye(numel(y0)));
```
Steven-Roberts, line 14: We should determine if the mass matrix needs to be sparse from the Jacobian.  Ideally, we could call `eye(n, 'like', J)` but this is not supported in Octave.

src/+otp/+utils/+solvers/dae34.m
```diff
@@ -0,0 +1,156 @@
...
+
+step = 1;
+
+while tc < tend
+
+    if tc + h > tend
+        h = tend - tc;
+    end
+
+    stages = zeros(numel(yc), stagenum);
```
Steven-Roberts, line 62: No need to reallocate every step

src/+otp/+utils/+solvers/dae34.m
```diff
@@ -0,0 +1,156 @@
...
+            stagedy = h*(stages(:, 1:(stage - 1))*A(stage, 1:(stage - 1)).');
+        else
+            stagedy = 0;
+        end
+
+        newtonk0 = zeros(size(yc));
+
+        np = inf;
+
+        ntol = 1e-6;
```
Steven-Roberts, line 80: This should be based on `AbsTol` and `RelTol`

src/+otp/+utils/+solvers/dae34.m
```diff
@@ -0,0 +1,156 @@
...
+        nmaxits = 5;
+        its = 0;
+        while norm(np) > ntol || its < nmaxits
+            Mc = M(staget, yc + stagedy + gh*newtonk0);
```
Steven-Roberts, line 84: If the mass matrix is state dependent, a Newton iteration would need dM/dy.  We should parse the `MStateDependence` property to handle this better

src/+otp/+utils/+solvers/dae34.m
```diff
@@ -0,0 +1,156 @@
...
+            g = Mc*newtonk0 - f(staget, yc + stagedy + gh*newtonk0);
+            H = Mc - gh*J(staget, yc + stagedy + gh*newtonk0);
+            [np, ~] = gmres(H, g);
```
Steven-Roberts, line 87: A direct solver would be more standard.  Unfortunately, Octave does not support `decomposition`, so `lu` is probably best.

src/+otp/+utils/+solvers/dae34.m
```diff
@@ -0,0 +1,156 @@
...
+
+        np = inf;
+
+        ntol = 1e-6;
+        nmaxits = 5;
+        its = 0;
+        while norm(np) > ntol || its < nmaxits
+            Mc = M(staget, yc + stagedy + gh*newtonk0);
+            g = Mc*newtonk0 - f(staget, yc + stagedy + gh*newtonk0);
+            H = Mc - gh*J(staget, yc + stagedy + gh*newtonk0);
```
Steven-Roberts, line 86: We can upgrade Newton iterations to algorithm on pg 119 of Solving ODEs II.  This is evaluation the Jacobian only as needed.

src/+otp/+utils/+solvers/dae34.m
```diff
@@ -0,0 +1,156 @@
...
+    end
+
+    yhat = yc + h*stages*bhat.';
+    yc = yc + h*stages*b.';
+
+    sc = abstol + max(abs(yc), abs(yhat))*reltol;
+
+    Mc = M(tc + h, yc);
+
+    err = rms((Mc*(yc-yhat))./sc);
```
Steven-Roberts, line 104: `rms` not supported in Octave

Steven-Roberts: Not sure why error is multiplied by `Mc`

src/+otp/+utils/+solvers/dae34.m
```diff
@@ -0,0 +1,156 @@
...
+
+if nargout < 2
+    sol = struct('x', t, 'y', y.');
+else
+    sol = t;
+end
+
+end
+
+function J = jacapprox(f, t, y)
```
Steven-Roberts, line 141: We'll need something a bit more robust or not support finite differences at all.

src/+otp/+utils/+solvers/dae34.m
AndreyAPopov: I can do manual rms. Error is multiplied by Mc, as one way to do error is to simply look at the differential variables and ignore the algebraic.

src/+otp/+utils/+solvers/dae34.m
AndreyAPopov: I found that the matrix is often very poorly scaled, so a direct solve was throwing tons of errors.

src/+otp/+utils/+solvers/dae34.m
AndreyAPopov: This is the only SDIRK (not ESDIRK) method with an embedded method that I could find with rational coefficients. Everything else is either not embedded or not rational (so single precision most of the time). If we have good coefficients I am all for it.

On order: are you sure it is order 2? So is the embedded method order 1?

AndreyAPopov: On index: I literally cannot find single method in the literature for index-3 problems. I think I have found a IRK method without an embedded method, but that's it. 

Again, if you can provide the coefficients, even if it is for IRK, we can work with it.

src/+otp/+utils/+solvers/dae43.m
```diff
@@ -0,0 +1,249 @@
+function [sol, y] = dae43(f, tspan, y0, options)
+
+n = numel(y0);
+
+h = odeget(options, 'InitialStep', []);
+reltol = odeget(options, 'RelTol', 1e-3);
+abstol = odeget(options, 'AbsTol', 1e-6);
+J = odeget(options, 'Jacobian', []);
+M = odeget(options, 'Mass', speye(numel(y0)));
```
Steven-Roberts, line 9: Default value should be `[]`

src/+otp/+utils/+solvers/dae43.m
```diff
@@ -0,0 +1,249 @@
...
+b = [1/6, 2/3, 1/6];
+c = [0, 1/2, 1];
+bhat = [-1/2, 2, -1/2];
+
+orderM = 4;
+orderE = 3;
+
+% Compute initial step size
+if isempty(h)
+    sc = abstol + reltol*abs(y0);
```
Steven-Roberts, line 45: `RelTol` could be a vector.  Should use `.*`

src/+otp/+utils/+solvers/dae43.m
```diff
@@ -0,0 +1,249 @@
...
+yc = y0;
+tc = tspan(1);
+
+stagenum = size(A, 1);
+tend = tspan(end);
+
+step = 1;
+
+Mc = M(tspan(1), y0);
+Mfull = zeros(n*stagenum, n*stagenum, 'like', Mc);
```
Steven-Roberts, line 73: Unfortunately, `'like'` option is not supported in Octave

src/+otp/+utils/+solvers/dae43.m
```diff
@@ -0,0 +1,249 @@
...
+
+if nargout < 2
+    sol = struct('x', t, 'y', y.');
+else
+    sol = t;
+end
+
+end
+
+function J = jacapprox(f, t, y)
```
Steven-Roberts, line 234: This seems like a useful enough to warrant it's own publicly-accessible function.  We can also look into more sophisticated and robust approaches.

src/+otp/+utils/+solvers/dae43.m
```diff
@@ -0,0 +1,249 @@
...
+end
+
+if isempty(M)
+    if usesparse
+        M = @(t, y) speye(numel(y0));
+    else
+        M = @(t, y) eye(numel(y0));
+    end
+elseif ~isa(M, 'function_handle')
+    M = @(t, y) M;
```
Steven-Roberts, line 31: We also need to support mass matrices that are only a function of time, e.g., `M(t)`

AndreyAPopov:
All resolved.

AndreyAPopov:
Should I remove the other DAE solvers (SDIRK and ESDIRK)?

Issue #40, opened 2023-04-12
Exact solutions


Steven-Roberts:
It would be nice to support exact solutions via a problem function `y = solveExactly(t)` which supports vector `t`. I see 2 approaches

1. Create an interface with a `solveExactly` function and only problems with exact solution implement it
2. Make `solveExactly` a `Problem` function that, by default, throws an errors that an exact solution is unavailable. Subclasses can override as needed.

I favor the latter. The list of problems with exact solutions is

- ascherlineardae?
- kpr
- linear
- protherorobinson
- trigonometricdae

PR #43, opened 2023-04-13
Add exact solutions

Steven-Roberts:
Added exact solutions for problems that have it. Most of these exact solutions only hold for a particular initial condition so I have to check that which is a bit fragile, but I don't see a better solution.

Addresses #40

### DIFF
--- a/src/+otp/+ascherlineardae/AscherLinearDAEProblem.m
+++ b/src/+otp/+ascherlineardae/AscherLinearDAEProblem.m
@@ -17,6 +17,16 @@ classdef AscherLinearDAEProblem < otp.Problem
                 'MStateDependence', 'none', ...
                 'MassSingular', 'yes');
         end
+        
+        function y = internalSolveExactly(obj, t)
+            beta = obj.Parameters.Beta;
+            if ~isequal(obj.Y0, [1; beta])
+                error('An exact solution is unavailable for this initial condition');
+            end
+            
+            y = [t .* sin(t) + (1 + beta * t) .* exp(-t); ...
+                beta * exp(-t) + sin(t)];
+        end
     end
 end
 
--- a/src/+otp/+kpr/KPRProblem.m
+++ b/src/+otp/+kpr/KPRProblem.m
@@ -24,5 +24,13 @@ classdef KPRProblem < otp.Problem
             obj.RHSSlow = otp.RHS(@(t,y) otp.kpr.fSlow(t, y, lambda, omega), ...
                 'Jacobian', @(t, y) otp.kpr.jacobianSlow(t, y, lambda, omega));
         end
+        
+        function y = internalSolveExactly(obj, t)
+            if ~isequal(obj.Y0, [2; sqrt(3)])
+                error('An exact solution is unavailable for this initial condition');
+            end
+            
+            y = sqrt([3 + cos(obj.Parameters.Omega * t); 2 + cos(t)]);
+        end
     end
 end
--- a/src/+otp/+linear/+presets/Dahlquist.m
+++ b/src/+otp/+linear/+presets/Dahlquist.m
@@ -2,15 +2,19 @@ classdef Dahlquist < otp.linear.LinearProblem
     %DAHLQUIST
     %
     methods
-        function obj = Dahlquist(A)
-            if nargin < 1 || isempty(A)
-                A = {-1};
-            end
-
+        function obj = Dahlquist(varargin)
             params = otp.linear.LinearParameters;
-            params.A = A;
             
-            obj = obj@otp.linear.LinearProblem([0, 1], ones(size(params.A{1}, 1), 1), params);
+            if nargin == 0
+                params.A = {-1};
+            else
+                params.A = varargin;
+            end
+            
+            A1 = params.A{1};
+            y0 = ones(size(A1, 1), 1, 'like', A1);
+            
+            obj = obj@otp.linear.LinearProblem([0, 1], y0, params);
         end
     end
 end
--- a/src/+otp/+linear/LinearProblem.m
+++ b/src/+otp/+linear/LinearProblem.m
@@ -40,5 +40,19 @@ classdef LinearProblem < otp.Problem
             obj.RHSPartitions = cellfun(@obj.createRHS, obj.Parameters.A, ...
                 'UniformOutput', false);
         end
+        
+        function y = internalSolveExactly(obj, t)
+            numT = length(t)
+            for i = 1:numT
+                yi = expm(t(i) * obj.RHS.Jacobian) * obj.Y0;
+                
+                if i == 1
+                    % The first yi provides the data type. Simply repeat vector for allocation
+                    y = repmat(yi, 1, numT);
+                else
+                    y(:, i) = yi;
+                end
+            end
+        end
     end
 end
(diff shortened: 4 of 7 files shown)


src/+otp/+linear/LinearProblem.m
```diff
@@ -40,5 +40,19 @@ function onSettingsChanged(obj)
             obj.RHSPartitions = cellfun(@obj.createRHS, obj.Parameters.A, ...
                 'UniformOutput', false);
         end
+        
+        function y = internalSolveExactly(obj, t)
+            numT = length(t)
+            for i = 1:numT
+                yi = expm(t(i) * obj.RHS.Jacobian) * obj.Y0;
```
AndreyAPopov, line 47: Shouldn't this technically be dependent on t(1) or on obj.TimeSpan(1)? I would do t(i) - obj.TimeSpan(1)

src/+otp/+protherorobinson/ProtheroRobinsonProblem.m
```diff
@@ -14,5 +14,13 @@ function onSettingsChanged(obj)
             obj.RHS = otp.RHS(@(t, y) otp.protherorobinson.f(t, y, lambda, phi, dphi), ...
                 'Jacobian', otp.protherorobinson.jacobian(lambda, phi, dphi));
         end
+        
+        function y = internalSolveExactly(obj, t)
+            if ~isequal(obj.Y0, obj.Parameters.Phi(obj.TimeSpan(1)))
+                error('An exact solution is unavailable for this initial condition');
```
AndreyAPopov, line 20: I would make this a standard error in the OTP namespace, so something like error('OTP:NoExactSolution', 'An exact...

src/+otp/+trigonometricdae/TrigonometricDAEProblem.m
```diff
@@ -13,6 +13,14 @@ function onSettingsChanged(obj)
                 'MassSingular', 'yes', ...
                 'Vectorized', 'on');
         end
+        
+        function y = internalSolveExactly(obj, t)
+            if ~isequal(obj.Y0, [sinh(0.5); tanh(0.5)])
+                error('An exact solution is unavailable for this initial condition');
```
AndreyAPopov, line 19: Same here

src/+otp/+trigonometricdae/TrigonometricDAEProblem.m
```diff
@@ -13,6 +13,14 @@ function onSettingsChanged(obj)
                 'MassSingular', 'yes', ...
                 'Vectorized', 'on');
         end
+        
+        function y = internalSolveExactly(obj, t)
+            if ~isequal(obj.Y0, [sinh(0.5); tanh(0.5)])
```
AndreyAPopov, line 18: Are we sure this will always work, maybe make sure it is within some epsilon?

src/+otp/+ascherlineardae/AscherLinearDAEProblem.m
```diff
@@ -17,6 +17,16 @@ function onSettingsChanged(obj)
...
                 'MassSingular', 'yes');
         end
+        
+        function y = internalSolveExactly(obj, t)
+            beta = obj.Parameters.Beta;
+            if ~isequal(obj.Y0, [1; beta])
+                error('An exact solution is unavailable for this initial condition');
+            end
+            
+            y = [t .* sin(t) + (1 + beta * t) .* exp(-t); ...
```
AndreyAPopov, line 27: I assume this also solves for the algebraic variable?

src/+otp/+kpr/KPRProblem.m
```diff
@@ -24,5 +24,13 @@ function onSettingsChanged(obj)
             obj.RHSSlow = otp.RHS(@(t,y) otp.kpr.fSlow(t, y, lambda, omega), ...
                 'Jacobian', @(t, y) otp.kpr.jacobianSlow(t, y, lambda, omega));
         end
+        
+        function y = internalSolveExactly(obj, t)
+            if ~isequal(obj.Y0, [2; sqrt(3)])
```
AndreyAPopov, line 29: sqrt(3) might not be exact, so approx equal? Is there a standard matlab function for equal to some precision?

src/+otp/+linear/+presets/Dahlquist.m
```diff
@@ -2,15 +2,19 @@
...
             
-            obj = obj@otp.linear.LinearProblem([0, 1], ones(size(params.A{1}, 1), 1), params);
+            if nargin == 0
+                params.A = {-1};
+            else
+                params.A = varargin;
+            end
+            
+            A1 = params.A{1};
+            y0 = ones(size(A1, 1), 1, 'like', A1);
```
AndreyAPopov, line 15: The way we are doing the splitting, I assume A1 is square in the size of y?

src/+otp/Problem.m
```diff
@@ -251,9 +265,8 @@ function validateNewState(obj, newTimeSpan, newY0, newParameters)
                 params = varargin(2:end);
             end
             
-            if ~(isvector(t) && otp.utils.validation.isNumerical(t))
-                error('The times must be a vector of numbers');
-            elseif ~(ismatrix(y) && otp.utils.validation.isNumerical(y))
+            t = obj.parseTime(t);
+            if ~(ismatrix(y) && otp.utils.validation.isNumerical(y))
                 error('The solution must be matrix of numbers');
```
AndreyAPopov, line 270: error('OTP:MustBeNumeric', '...')

src/+otp/Problem.m
```diff
@@ -268,5 +281,13 @@ function validateNewState(obj, newTimeSpan, newY0, newParameters)
                 error('Expected solution to have %d variables but has %d', obj.NumVars, n);
             end
         end
+        
+        function t = parseTime(obj, t)
+            if ~isvector(t) || isempty(t) || ~otp.utils.validation.isNumerical(t)
+                error('The times must be a nonempty vector of numbers');
```
AndreyAPopov, line 287: I don't like the naming of t, maybe 'The input argument must be...'

src/+otp/+protherorobinson/ProtheroRobinsonProblem.m
Steven-Roberts: All errors should have this I think. I'll save it for another PR, though, and just made issue about it.

src/+otp/+trigonometricdae/TrigonometricDAEProblem.m
Steven-Roberts: This will probably fail for non-double data types, but at least it is always conservative. If the user never modifies the IC from default it should be fine. I hesitate to allow tolerances cause you can always manufacture edge cases where things break.

src/+otp/+kpr/KPRProblem.m
Steven-Roberts: This one will be tricky for VPA and gpuArray, but I think that's out of scope for now. Honestly, I think it's good enough for now. Surprisingly it handles cases like `isequal(sqrt(3), single(sqrt(3)))`

src/+otp/Problem.m
Steven-Roberts: Not sure I understand. This function also gets called when solution struct (and not explicit t argument) is passed to `problem.solve`.

AndreyAPopov: Hmmm, so it is multi-purpose. OK that makes it harder to change. It is fine for now.

AndreyAPopov:
Might be ready to merge in the current state

PR #45, opened 2023-04-16
Add err ID to all errors

Steven-Roberts:


src/+otp/+ascherlineardae/AscherLinearDAEProblem.m
```diff
@@ -21,7 +21,8 @@ function onSettingsChanged(obj)
         function y = internalSolveExactly(obj, t)
             beta = obj.Parameters.Beta;
             if ~isequal(obj.Y0, [1; beta])
-                error('An exact solution is unavailable for this initial condition');
+                error('OTP:noExactSolution', ...
```
AndreyAPopov, line 24: It seems like there is no consistent casing, and lower case 'no' has precedent in matlab. Weird, but this seems correct.

src/+otp/+utils/+movie/+recorder/NullVideoRecorder.m
```diff
@@ -26,7 +26,7 @@ function stop(~)
         
         function h = play(~)
             h = 'Video not saved for playback';
-            error(h);
+            error('OTP:movieNotSaved', h);
```
AndreyAPopov, line 29: maybe do 'OTP:noVideoSaved'? or 'OTP:videoNotSaved'? We use the term video and not movie.

src/+otp/+utils/+movie/CometMovie.m
```diff
@@ -54,19 +50,16 @@ function drawFrameAxes(obj, ~, gObjects, state)
...
         function x = getXPoints(obj, cometIdx, state)
             x = otp.utils.compatibility.abstract(obj, cometIdx, state);
         end
         
         function y = getYPoints(obj, cometIdx, state)
             y = otp.utils.compatibility.abstract(obj, cometIdx, state);           
         end
+        
+        function z = getZPoints(obj, cometIdx, state)
+            z = otp.utils.compatibility.abstract(obj, cometIdx, state);
```
AndreyAPopov, line 62: Does 3d work now?

AndreyAPopov:
Everything else other than some minor comments looks good

src/+otp/+utils/+movie/+recorder/NullVideoRecorder.m
Steven-Roberts: Fixed in other PR so movie is always used

PR #46, opened 2023-04-16
Remove video from recorder names

Steven-Roberts:
Eliminates inconsistent use of movie vs video


AndreyAPopov:
So we are still calling it video, right?

AndreyAPopov:
Looks good otherwise.

Steven-Roberts:
> So we are still calling it video, right?

No we use movie everywhere now

PR #47, opened 2023-04-19
Use prop validation for Problem and remove validateNewState

Steven-Roberts:


src/+otp/Problem.m
```diff
@@ -1,67 +1,95 @@
...
     
     properties (SetAccess = private)
-        Name % A human-readable representation of the problem
+        % A human-readable representation of the problem
+        Name %MATLAB ONLY: (1,:) char
     end
     
     properties (SetAccess = protected)
-        RHS
+        RHS %MATLAB ONLY: (1,1) otp.RHS = otp.RHS(@() [])
```
AndreyAPopov, line 9: What does this do?

src/+otp/Problem.m
```diff
@@ -1,67 +1,95 @@
...
     end
     
     properties (Access = private)
-        Settings
-        ExpectedNumVars
+        ExpectedNumVars %MATLAB ONLY: {mustBeScalarOrEmpty, mustBeInteger, mustBePositive}
+        
+        % Determines if the Problem properties are still being set. Once set, it
+        % is ok to call onSettingsChanged.
+        Initialized = false
```
AndreyAPopov, line 17: Oh, so now we won't call onsettingschanged three times when making the problem?

Steven-Roberts: Exactly. This was true before this PR but handled differently

src/+otp/Problem.m
```diff
@@ -1,67 +1,95 @@
...
         
         function timeSpan = get.TimeSpan(obj)
-            timeSpan = obj.Settings.timeSpan;
+            timeSpan = obj.InternalTimeSpan;
         end
         
         function set.Y0(obj, value)
-            obj.Settings.y0 = value;
+            numVars = length(value);
+            if ~isempty(obj.ExpectedNumVars) && numVars != obj.ExpectedNumVars
```
AndreyAPopov, line 67: "!=" should be "~=" this fails in matlab

AndreyAPopov:
All tests run correctly after changing the != to ~=


Then for brusselator:

`>> m.Y0 = [3; 1; 3]
Error using otp.Problem/set.Y0
Expected Y0 to have 2 components but has 3`

`>> m.TimeSpan = [0, 1, 2];
Error setting property 'TimeSpan' of class 'otp.Problem'. Value must be a vector with
2 elements.
`

So that works. Everything else seems to work.

PR #48, opened 2023-04-19
Added Quick Start Notebook

elswit:
to configure JupyterHub servers for the Ocatave demo notebook


elswit (commented):
quick fix for the path directory

PR #49, opened 2023-04-21
Document octave bugs

Steven-Roberts:
Let me know if there are any other octave-specific issues


src/+otp/+bouncingball/onEvent.m
```diff
@@ -5,8 +5,9 @@
 % horizontal plane, we then invert the y velocity, and transform
 % back.
 
-pos = sol.ye(1:2, end);
-vel = sol.ye(3:4, end);
+% OCTAVE BUG: event data is transposed compared to MATLAB
```
AndreyAPopov, line 8: Is it possible to do some type of try catch thing to do different stuff for matlab and octave here so that everything works?

src/+otp/Problem.m
```diff
@@ -149,7 +149,8 @@
     end
     
     methods (Access = protected)
-        % This method is called when either TimeSpan, Y0, or parameters are changed.  It should update F and other properties such as a Jacobian to reflect the changes.  This is effevtively an abstract function but not explicitly marked so in order to support Octave
+        % This method is called when either TimeSpan, Y0, or parameters are changed. It should
```
AndreyAPopov, line 152: What max line width are we defaulting to? I personally like 120 instead of 80 (we are not in the 70s anymore), but your call really.

src/+otp/+qg/f.m
```diff
@@ -6,6 +6,7 @@
 psi = reshape(psi, nx, ny, []);
 
 % Calculate the vorticity
+% OCTAVE BUG: pagemtimes not supported
 q = -(pagemtimes(Lx, psi) + pagemtimes(psi, Ly));
```
AndreyAPopov, line 10: Is there a way to define a function ONLY if we are in octave?

src/+otp/+bouncingball/onEvent.m
Steven-Roberts: The reshape handles things correctly for both MATLAB and Octave. I'll label this as `OCTAVE FIX` instead.

src/+otp/Problem.m
Steven-Roberts: 120 seems reasonable for the code. My only minor concern is when user calls `help` in a narrow command window it wrap doc text in ugly way.

src/+otp/+utils/Solver.m
```diff
@@ -1,12 +1,12 @@
 classdef Solver
     properties (Constant)
         Nonstiff = @ode45        
-        Stiff = otp.utils.compatibility.isOctave(@ode23s, @ode15s) % OCTAVE BUG: ode15s does not
+        Stiff = otp.utils.compatibility.isOctave(@ode23s, @ode15s) % OCTAVE FIX: ode15s does not
```
AndreyAPopov, line 4: Can I use something similar to fix QG pagemtimes?

Steven-Roberts: Yes. Could even have a separate f function. Unless it's a simple fix, I don't think it's necessary to fix in this PR

AndreyAPopov: Not this PR, but I am downloading octave to test for QG

src/+otp/RHS.m
```diff
@@ -39,6 +39,12 @@
                 f = fields{i};
                 obj.(f) = extras.(f);
             end
+            
+            % OCTAVE FIX: ode15s throws an error when the Vectorized option is set. It seems no
+            % integrator uses this option anyway.
+            if otp.utils.compatibility.isOctave()
+                obj.Vectorized = [];
```
AndreyAPopov, line 46: In octave or in matlab? No integrator uses this option?

Steven-Roberts: Octave only

src/+otp/+qg/f.m
AndreyAPopov: This is now fixed in my new QG pull request, so you can revert this file

AndreyAPopov:
CUSP, lorenz96, and nbody took a super long time to solve in octave. ASCHERLINEARDAE also fails, but this is expected right?

Steven-Roberts:
The performance difference is mostly due to using `ode23s` in Octave instead of `ode15s`. How is Ascher problem failing? Works fine for me.

AndreyAPopov:
`model.solve
error: @<anonymous>: function called with too many inputs
error: called from
    onSettingsChanged>@<anonymous>
    ode23s at line 203 column 18
    internalSolve at line 214 column 17
    solve at line 110 column 17
>>`

Steven-Roberts:
Hmm what version of Octave?

AndreyAPopov:
8.2.0

AndreyAPopov:
It looks like in my version of octave, ode23s expects the mass function to have two arguments, t and x, but ascherlineardae only provides Mass(t)

Steven-Roberts:
This PR changes the default solver for Ascher problem to ode15s so this shouldn't be a problem

AndreyAPopov:
I was on the octave branch not the octave_compat branch ...

AndreyAPopov:
Bouncingball should still fail right?

AndreyAPopov:
I pushed a fix for bouncing ball. It will fail in octave unless MaxStep is set to 1e-2 or lower, as octave has a bug that makes it fail to detect some events (thus causing the ball to go below the ground).

AndreyAPopov:
All problems now build and solve in octave. Hooray!

src/+otp/+utils/+compatibility/odextend.m
```diff
@@ -0,0 +1,44 @@
+function solExtend = odextend(sol, f, tfinal, y0, opts)
+% OCTAVE FIX: naive odextend implementation to fix event function
```
AndreyAPopov, line 2: Oh wow, nice fix for my naive implementation. I know you didn't want to implement this, but I feel like having all problems work with octave would be super nice.

Very nice.

src/+otp/Problem.m
```diff
@@ -302,7 +296,7 @@ function onSettingsChanged(obj)
             end
         end
         
-        function t = parseTime(obj, t)
+        function t = parseTime(~, t)
```
AndreyAPopov, line 299: If parseTime does not depend on the object, then it is a static function right? so we can make it static.

Steven-Roberts:
I think we're ready to merge, right?

AndreyAPopov:
Basically, unless you feel like change the parseTime thing.

PR #52, opened 2023-04-25
Document Brusselator problem

Steven-Roberts:
This will establish conventions for documenting other problems.

### DIFF
--- a/src/+otp/+brusselator/BrusselatorParameters.m
+++ b/src/+otp/+brusselator/BrusselatorParameters.m
@@ -1,9 +1,10 @@
 classdef BrusselatorParameters
-    %BRUSSELATORPARAMETERS 
+    %BRUSSELATORPARAMETERS Parameters for the Brusselator problem
+    %   See also otp.brusselator.BrusselatorProblem
     properties
-        %A is a combination of an excess reactant and reaction rate
+        %A concentration of excess reactant A in the Brusselator reaction
         A %MATLAB ONLY: (1,1) {mustBeNumeric}
-        %B is a combination of an excess reactant and reaction rate
+        %B concentration of excess reactant B in the Brusselator reaction
         B %MATLAB ONLY: (1,1) {mustBeNumeric}
     end
 end
--- a/src/+otp/+brusselator/BrusselatorProblem.m
+++ b/src/+otp/+brusselator/BrusselatorProblem.m
@@ -21,7 +21,7 @@ classdef BrusselatorProblem < otp.Problem
     %       Number of Variables: 2
     %       Stiff: no
     %
-    %   Bibliography:
+    %   Sources:
     %   Lefever, R., and G. Nicolis. "Chemical Instabilities and Sustained Oscillations." Journal of Theoretical
     %   Biology, vol. 30, no. 2, Elsevier BV, Feb. 1971, pp. 267–284. https://doi.org/10.1016/0022-5193(71)90054-3.
     %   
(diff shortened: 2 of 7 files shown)


src/+otp/+brusselator/BrusselatorParameters.m
```diff
@@ -1,9 +1,10 @@
 classdef BrusselatorParameters
-    %BRUSSELATORPARAMETERS 
+    %BRUSSELATORPARAMETERS Parameters for the Brusselator problem
+    %   See also otp.brusselator.BrusselatorProblem
     properties
-        %A is a combination of an excess reactant and reaction rate
+        %A concentration of excess reactant A in the Brusselator reaction
```
AndreyAPopov, line 5: this is not true, because it also represents the reaction rate in one of the directions.

The way the problem is discretized and simplified, the reaction rate and concentration of excess reactant are combined into one dimensionless constant.

same comment for B

src/+otp/+brusselator/BrusselatorProblem.m
```diff
@@ -1,34 +1,55 @@
...
+    %BRUSSELATORPROBLEM A two-variable model for an autocatalytic reaction
+    %   The Brusselator chemical reaction is given by
     %
     %        A → X
     %    B + X → Y + D
     %   2X + Y → 3X
     %        X → E
     %
-    %   This system can be reduced to the following two equations
+    %   With the assumption that all reaction rates are one and the concentrations of A and B are constant parameters,
```
AndreyAPopov, line 10: I am not sure that this is true either. we are simply reducing the number of reactions to two and non-dimensionalizing the problem.

AndreyAPopov:
I like the documentation style, I am just concerned about the veracity of some of the things you say. Is there a source that you got this from? Because it does not seem correct, unless I am having a brain fart moment.

AndreyAPopov:
Ok,

so A is concentration times k and B is concentration times a different k, nothing is non-dimensionalized. 

So A is technically in dimension concentration squared divided by time, and not really the concentration, but we can treat the parameter as just representing the concentration, that is implicitly then multiplied by a unit reaction rate.

I have convinced myself.

src/+otp/+brusselator/BrusselatorProblem.m
```diff
@@ -1,34 +1,55 @@
...
+    %
+    %   Problem Properties:
+    %       Type: ODE
+    %       Number of Variables: 2
+    %       Stiff: no
     %
     %   Bibliography:
-    %   Ault, Shaun, and Erik Holmgreen. "Dynamics of the Brusselator."
-    %   Math 715 Projects (Autumn 2002) (2003): 2.
+    %   Lefever, R., and G. Nicolis. "Chemical Instabilities and Sustained Oscillations." Journal of Theoretical
```
AndreyAPopov, line 25: Do we have a preferred citation style, or anything goes?

Steven-Roberts: MLA
https://github.com/ComputationalScienceLaboratory/ODE-Test-Problems/blob/master/docs/CONTRIBUTING.md#documentation

Steven-Roberts:
Most of the description comes from Solving ODEs I pg 115-116.

src/+otp/+brusselator/BrusselatorProblem.m
AndreyAPopov: Oh perfect, I was going to suggest adding it to Contributing.md, but looks like you have your bases covered, fantastic!

src/+otp/+brusselator/BrusselatorProblem.m
```diff
@@ -21,7 +21,7 @@
     %       Number of Variables: 2
     %       Stiff: no
     %
-    %   Bibliography:
+    %   Sources:
```
AndreyAPopov, line 24: Maybe references? Sources can be confused with source code?

elswit (approved):

PR #57, opened 2023-04-30
Testing enhancement

AndreyAPopov:
I am in the process of reworking the tests and adding more.

The derivative code is very much a work-in-progress, so it is not much to look at at the moment.


src/+otp/+utils/+derivatives/jacobian.m
```diff
@@ -0,0 +1,76 @@
...
+
+end
+
+function J = adjacobianinternal(f, t, y)
+
+n = numel(y);
+
+fy = f(t, y);
+
+J = zeros(n, n, 'like', y);
```
Steven-Roberts, line 70: For Octave compatibility, you could use the backward iteration trick to avoid explicit allocation.

src/tests/validatederivatives.m
```diff
@@ -0,0 +1,205 @@
...
+        continue;
+    end
+
+    % setup
+    tc = model.TimeSpan(1);
+    y0 = model.Y0;
+    f = model.RHS.F;
+
+    % Try to see if the Jacobian works
+    if ~isempty(model.RHS.Jacobian)
```
Steven-Roberts, line 34: `RHS.JacobianFunction` should be able to simply the logic

Steven-Roberts (changes requested):

src/+otp/+utils/+derivatives/jacobian.m
AndreyAPopov: Octave does not support dlarray, so I don't think that is needed.

src/+otp/+utils/+derivatives/jacobianadjointvectorproduct.m
```diff
@@ -0,0 +1,87 @@
+function Javp = jacobianadjointvectorproduct(f, t, y, u, method)
```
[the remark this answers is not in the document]
AndreyAPopov: Yeah, that's a good idea. I might refactor as a class, I think that would make a lot of sense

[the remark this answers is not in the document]
Steven-Roberts: I was thinking a single function with extra option for adjoint. The full Jacobian is just JVP with identity matrix.

src/+otp/+utils/+derivatives/jacobian.m
Steven-Roberts: Ah true

src/+otp/+utils/+derivatives/jacobianadjointvectorproduct.m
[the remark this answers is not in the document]
AndreyAPopov: I will just do a jacobian for now. For testing that is all that we need.

AndreyAPopov:
Alright, this seems to be in a good place now. It uses asserts, and fails if anything wrong happens. I also optimised the runtime.

Steven-Roberts:
A simpler way to load all the presets:
```
function presets = getPresets()

presets = arrayfun(@constructPreset, dir('+otp/*/+presets/*.m'), 'UniformOutput', false);

end

function preset = constructPreset(file)

[~, presetName, ~] = fileparts(file.name);
folders = strsplit(file.folder, filesep)
problemName = folders{end-1};
preset = eval(sprintf('otp.%s.presets.%s', problemName(2:end), presetName));
    
end
```

AndreyAPopov:
> A simpler way to load all the presets:
> 
> ```
> function presets = getPresets()
> 
> presets = arrayfun(@constructPreset, dir('+otp/*/+presets/*.m'), 'UniformOutput', false);
> 
> end
> 
> function preset = constructPreset(file)
> 
> [~, presetName, ~] = fileparts(file.name);
> folders = strsplit(file.folder, filesep)
> problemName = folders{end-1};
> preset = eval(sprintf('otp.%s.presets.%s', problemName(2:end), presetName));
>     
> end
> ```

This doesn't work in general. I'm trying to generalize it to when the working dir is not src

elswit (approved):

PR #61, opened 2023-05-02
Notebooks

elswit:
Adds ``gitattributes`` to fix the language detection bug introduced after using ``.ipynb`` files. Also updates Jupyter notebook for quick-start to include a table of current test problems


Steven-Roberts:
Looks like some extra files got committed by accident

PR #66, opened 2023-05-07
Documentation using sphinx

AndreyAPopov:


Steven-Roberts:
Docs look great! Love the organization and search functionality. Here's some thoughts:
- To avoid committing all these generated files can we just have GitHub actions taken care of all doc building and publishing steps?
- Some public functions in problem subclasses (e.g. for postprocessing) are not shown 
- We'll need to add pages for RHS and problem. Let's skip utils for now 
- It would be nice to have a single readme.rst that works as GitHub readme and docs homepage to avoid duplication 
- The current theme is not mobile friendly

AndreyAPopov:
> * To avoid committing all these generated files can we just have GitHub actions taken care of all doc building and publishing steps?

Yes, I think Arash will look into it. It is possible, but not right now.

> * Some public functions in problem subclasses (e.g. for postprocessing) are not shown

I think this is an issue of them actually lacking documentation. If they don't have a comment on the next line after being defined, the docs won't show the function.

> * We'll need to add pages for RHS and problem. Let's skip utils for now

Yes.

> * It would be nice to have a single readme.rst that works as GitHub readme and docs homepage to avoid duplication

That might be difficult, but we can try.

> * The current theme is not mobile friendly

Oh, I didn't even try that. I just used the classic theme as it is the most customizable. There is also the "Read-the-docs" theme, which is externally maintained as SUPER popular. That might be our best bet.

Steven-Roberts:
Yay the site is being built with an action! I think with a final review, this can be merged.

AndreyAPopov:
I read through everything.

The modern python way of doing strings looks really nice. 

The site seems to be working too.

Very nice!

Let's merge?

PR #92, opened 2023-06-02
Readme patch

elswit:


elswit:
an interesting side effect here is that docu buid test checks out while deploy fails because branch of the pr is still not master. Have to work on this.

PR #82, opened 2023-05-20
Add tests for plot and movie and QG

AndreyAPopov:

PR #93, opened 2023-06-05
Feature/gifs

elswit:
added script that lists all test problems and creates movies for them


gallery/makeMovie.m
```diff
@@ -0,0 +1,84 @@
+close all
+
+presets = getpresets();
+
+if ~ isfolder('animations')
```
Steven-Roberts, line 5: This check isn't needed. `mkdir` doesn't throw an error if it already exists

gallery/makeMovie.m
```diff
@@ -0,0 +1,84 @@
...
+[indx,tf] = listdlg('PromptString', ...
+    {'Select a test problem',...
+    'Only one problem can be selected at a time.',''},...
+    'ListSize',[250,500], ...
+    'InitialValue', 22, ...
+    'OKString','Select', ... 
+    'SelectionMode','single','ListString',listAllExperiments);
+
+if tf
+    eval(strcat(['problem =', listAllExperiments{indx}]));
```
Steven-Roberts, line 26: `problem = eval(listAllExperiments{indx})`

gallery/makeMovie.m
```diff
@@ -0,0 +1,84 @@
...
+if ~ isfolder('animations')
+    mkdir('animations')
+end
+
+
+listAllExperiments = {};
+
+for preset = presets
+    listAllExperiments{end+1} = preset.presetclass;
+end
```
Steven-Roberts, line 14: listAllExperiments = {preset.presetclass};

gallery/makeMovie.m
```diff
@@ -0,0 +1,84 @@
...
+    end
+    
+
+    % convert vido format if neccessary 
+    switch vid_format
+        case 'Avi'
+            
+        case 'Gif'
+    
+            args     = strcat(['-i ' ...
```
Steven-Roberts, line 47: No need to call `strcat` if you have square brackets

gallery/makeMovie.m
```diff
@@ -0,0 +1,84 @@
...
+            filename,'.webm']);
+
+    end
+
+    if ~strcmp(vid_format,'Avi') && ~isempty(vid_format)
+
+        prompt = {'Ffmpeg is required for video conversion'};
+        dlgtitle = 'Enter path to ffmpeg:';
+        dims = [1 35];
+        definput = {'/opt/homebrew/bin/ffmpeg'};
```
Steven-Roberts, line 65: I suggest making this file a function with this an an argument. By default this can be `ffmpeg` which assumes it's in the path

Steven-Roberts (changes requested):
This very general functionality fits well with the movie tests in #82 btw.

Here's another idea: what if we generate a movie for all problems (or presets) and include it in the sphinx docs. That would look cool online.

elswit:
I've updated the script, basically getting rid of all the gui since we will only use the animation script internally.

elswit (commented):
Moved the plotting script to `images` subdirectory to avoid cluttering

PR #95, opened 2023-06-06
Follow MATLAB Toolbox Best Practices

Steven-Roberts:
This may cause merge issues with current PRs


Steven-Roberts:
I think we should merge this ASAP after review so we commit to directory structure.

PR #99, opened 2023-06-23
HTML table in readme.md to put  logo and text side by side 

elswit:
See if you like this way of embedding the logo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 
 
-<table>
+<table border="0";>
 <tbody>
     <tr>
         <td> <img src="images/logo.png" width="5000" alt='ODE Test problems Logo' /></td>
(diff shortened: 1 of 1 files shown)


elswit:
See the new readme here, the diff doesn't show much

https://github.com/ComputationalScienceLaboratory/ODE-Test-Problems/blob/898b8f1a4e07cb68009048abfc0ca1b98547e0c6/README.md

Steven-Roberts:
On mobile, the columns are too narrow. Once we have gifs for readme we can see about different image layouts.

elswit:
> On mobile, the columns are too narrow. Once we have gifs for readme we can see about different image layouts.

 I'll merge this. I'll  add the gifs in a separate branch and we'll experiment with the layouts there

Issue #64, opened 2023-05-03
Make parameters inherit a base parameter class


AndreyAPopov:
The class hierarchy would make more sense if this was the case.

Steven-Roberts:
Is there any shared functionality the base class would provide? If not, I question if it's necessary

AndreyAPopov:
It would help clean up the class hierarchy in the documentation. We can talk about it offline.

AndreyAPopov:
I don't think this is needed anymore. Closing

PR #117, opened 2023-09-06
Docs/ascher DAE

elswit:


elswit:
Note: I'm not using parameter class #64 yet.

PR #128, opened 2024-04-15
Add docs to linear problem

Steven-Roberts:


docs/conf.py
```diff
@@ -32,6 +32,16 @@
     'show-inheritance': True
 }
 
+mathjax3_config = {
+    'tex': {
+        'macros': {
+            'diag': ['\\operatorname{diag}']
```
AndreyAPopov, line 38: Oh very nice that we can add macros!

docs/conf.py
```diff
@@ -32,6 +32,16 @@
...
+        }
+    }
+}
+
+myst_heading_anchors = 2
```
AndreyAPopov, line 43: What does this do?

toolbox/+otp/+linear/+presets/Alpha.m
```diff
@@ -1,8 +1,27 @@
 classdef Alpha < otp.linear.LinearProblem
-    %ALPHA
+    % A diagonal linear preset with eigenvalues logarithmically spaced between distances $r_{min}$ and $r_{max}$ from
+    % the origin at an angle $α$ measured in degrees clockwise from the negative real axis. It can be used to check for
+    % $A(α)$ stability of a time integration scheme. This preset uses $t ∈ [0, 1]$, $y_0 = [1, 1, …, 1]^T$, and
     %
+    % $$
+    % Λ_1 &= \diag(-r_1 e^{-i π α / 180}, -r_2 e^{-i π α / 180}, …, -r_N e^{-i π α / 180}), \\
+    % r_j &= r_{min}^{\frac{N - j}{N - 1}} r_{max}^{\frac{j - 1}{N - 1}}.
```
AndreyAPopov, line 8: can we do \max and \min ? that might make it look better

docs/conf.py
Steven-Roberts: Fixes an unrelated warning from the contributing guide markdown

PR #129, opened 2024-04-21
Add docs to Problem

Steven-Roberts:


toolbox/+otp/+utils/+movie/+recorder/MemoryRecorder.m
```diff
@@ -29,8 +29,12 @@ function stop(~)
             % Nothing to do
         end
         
-        function h = play(obj)
-            h = implay(obj.Mov, obj.FrameRate);
+        function play(obj)
+            if exist('implay', 'builtin')
```
Steven-Roberts, line 33: TODO: check this in MATLAB

Steven-Roberts, line 33:
```suggestion
            if exist('implay', 'file')
```

